PR #8, opened 2022-07-11
Common Components

### LINKED ISSUE #5: Common components
- [x] Button
- [x] Input
- [x] Base styles
- [x] Frontend folder structuring

sandje00:
THIS PR:

- Base frontend folder structure
- Base button
- Base input
- Base field
- Base base textarea

Related issues:
Closes #5

### DIFF
--- a/frontend/src/App.vue
+++ b/frontend/src/App.vue
@@ -2,6 +2,17 @@
   <div></div>
 </template>
 
-<script></script>
+<script>
+export default {
+  name: 'app-main',
+};
+</script>
 
-<style></style>
+<style>
+@import './assets/stylesheets/main.css';
+
+#app {
+  width: 100%;
+  height: 100%;
+}
+</style>
--- a/frontend/src/App.vue
+++ b/frontend/src/App.vue
@@ -9,7 +9,8 @@ export default {
 </script>
 
 <style>
-@import './assets/stylesheets/main.css';
+@import 'normalize.css';
+@import '@/assets/stylesheets/main.css';
 
 #app {
   width: 100%;
--- a/frontend/src/App.vue
+++ b/frontend/src/App.vue
@@ -11,9 +11,4 @@ export default {
 <style>
 @import 'normalize.css';
 @import '@/assets/stylesheets/main.css';
-
-#app {
-  width: 100%;
-  height: 100%;
-}
 </style>
--- a/frontend/src/components/common/ChiButton.vue
+++ b/frontend/src/components/common/ChiButton.vue
@@ -1,5 +1,5 @@
 <template>
-  <button class="button">
+  <button class="chi-button">
     <slot></slot>
   </button>
 </template>
@@ -11,7 +11,7 @@ export default {
 </script>
 
 <style scoped>
-.button {
+.chi-button {
   background: none;
   border: none;
   cursor: pointer;
--- a/frontend/src/components/common/ChiButton.vue
+++ b/frontend/src/components/common/ChiButton.vue
@@ -6,7 +6,7 @@
 
 <script>
 export default {
-  name: 'v-button',
+  name: 'chi-button',
 };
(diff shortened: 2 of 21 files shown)


frontend/src/components/common/vTextarea.vue
```diff
@@ -0,0 +1,27 @@
...
+  },
+};
+</script>
+
+<style>
+.textarea {
+  font-size: inherit;
+  background: none;
+  outline: none;
+  resize: none;
```
Lnola, line 25: remove border

Lnola (changes requested):
Needs a few simple changes, nothing major

frontend/src/components/common/vButton.vue
```diff
@@ -0,0 +1,19 @@
+<template>
+  <button class="button">
```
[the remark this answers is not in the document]
sandje00: It is not necessary, since `click` event handling will be defined in a context of a component in which base button is used.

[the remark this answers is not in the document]
Lnola: react mentality :)

Lnola (approved):
Looks good!

frontend/src/App.vue
```diff
@@ -2,6 +2,18 @@
...
+  name: 'app-main',
+};
+</script>
 
-<style></style>
+<style>
+@import 'normalize.css';
+@import '@/assets/stylesheets/main.css';
+
+#app {
```
abasic, line 15: Please move this to the `global.css` file.

frontend/src/components/common/ChiButton.vue
```diff
@@ -0,0 +1,19 @@
+<template>
+  <button class="button">
```
abasic, line 2: Please use more specific class names.

frontend/src/components/common/ChiButton.vue
```diff
@@ -0,0 +1,19 @@
...
+  </button>
+</template>
+
+<script>
+export default {
+  name: 'chi-button',
+};
+</script>
+
+<style>
```
abasic, line 13: Please use the `scoped` attribute if appropriate.

frontend/src/composables/useModelValue.js
```diff
@@ -0,0 +1,7 @@
+export default function useModelValue(emit) {
```
abasic, line 1: Please use the `v-bind="$attrs"` approach if applicable.

abasic (changes requested):
Some of the comments can be applied to more than the line they were written on. Please apply them everywhere they are applicable.

frontend/src/components/common/ChiInput.vue
```diff
@@ -0,0 +1,30 @@
+<template>
+  <input v-bind="{ ...$attrs, onInput: updateModelValue }" :value="modelValue" class="chi-input" />
```
abasic, line 2: The code I had in mind when I mentioned `$attrs` being used is this:
```vue
<input v-bind="$attrs" class="chi-input">
```
I believe this would achieve the same desired behavior. The reason is that a parent component could use it in the same way as a regular input tag. So `chi-input` and `input` could be interchangeable.
```vue
<input v-model="xyz">
<chi-input v-model="abc">
```
and
```vue
<input @input="update($event)" :value="xyz">
<chi-input @input="update($event)" :value="abc">
```

I am not certain if the way it is written now could be used interchangeably. Looking at it again, I am not sure a `v-model` would work correctly. Interchangeability in this case, to me, is a worthwhile goal because it makes the interface of the custom component the same as the native one and it is more straightforward to use.

Please take a look and see if what I wrote makes sense and is in line with your goals.

abasic: `v-model` was changed in Vue 3 so this does not work.

PR #14, opened 2022-07-13
Update docker entrypoint

Lnola:
Updated docker entrypoint to allow installation on packages without having to drop containers 📦


sandje00 (approved):
Looks good ✅

PR #19, opened 2022-07-18
Authorization

### LINKED ISSUE #15: Authorization

sandje00:
THIS PR:

- Add `refreshToken` column to `users` table (create new migration and update model)
- Add access and refresh token generation abstraction in form of user instance method
- Add access and refresh token generation method calls inside `login` and `register` user controller functions
- Send access and refresh token to frontend inside cookies (access token is a regular and refresh token is an HTTP only cookie)
- Extend Express Request interface with user
- Add `authenticate()` as access token validation and user authentication middleware function
- Add `refresh()` as refresh token validation and access token issuing middleware function

The implementation of user authorisation using access and refresh token in this PR is basic; refresh token is saved in a dedicated column of user table. More complex implementations have been left for the future ad bonus feature.

Also, `authenticate()` and `refresh()` middleware functions are meant to be used one after another in express pipeline. Example:

```
router
  .use(authenticate)
  .use(refresh)
  .get('/', getAll);
```
or
```
router.get('/', authenticate, refresh, getAll);
```

Closes #15


backend/src/shared/auth/refresh.ts
```diff
@@ -0,0 +1,37 @@
...
+  const { refreshToken } = req.cookies;
+  if (!refreshToken) {
+    return res.status(403).json({ error: FORBIDDEN_ERROR });
+  }
+  try {
+    const user = await User.findOne({ where: { refreshToken } });
+    const { id, username, aud } = jwtVerify(refreshToken, process.env.REFRESH_TOKEN_SECRET || '') as IJwtPayloadDecoded;
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+    // @ts-ignore
+    if (!user || id !== user.id || username !== user.username || aud !== Audience.Scope.Refresh) {
```
sandje00, line 19: I am aware this is just a hack rather than a clean solution; I would like to replace it with the better solution of the problem, but since finding it took far too much of my time, I decided to avoid the problem temporarily this way to finish the functionality and get the rest of the code ready for the review.

THE PROBLEM:
Even though Express Request interface is extended (`@types/express/index.d.ts`) with `user` property of type `User` (which is a user model), inside `refresh` middleware function TS recognises fetched user as a Sequelize model, but not as user model instance specifically, which means that TS forbids accessing instance data properties and instance methods.

Inside `@types/express/index.d.ts` I have tried to set type of `user` property:

- custom `IUser` interface that exists on branch main
- user model that implements `IUser` interface that exists on branch main
- user model newly rewritten in accordance with (Sequelize documentation)[https://sequelize.org/docs/v6/other-topics/typescript/]
- writing all the data properties and instance methods manually.

The only of those four that worked was the last one, but it would be much better to leverage what has already been defined, i. e. something like the first three.

I am open to any ideas that would help to solve the problem.

backend/src/shared/auth/refresh.ts
```diff
@@ -0,0 +1,37 @@
...
+      return res.status(403).json({ error: FORBIDDEN_ERROR });
+    }
+    req.user = user;
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+    // @ts-ignore
+    const { accessToken: newAccessToken, refreshToken: newRefreshToken } = await user.generateTokens();
```
sandje00, line 25: The same as the previous comment.

backend/src/user/user.model.ts
```diff
@@ -1,13 +1,18 @@
...
-  username!: string;
-  password!: string;
+// eslint-disable-next-line no-use-before-define
+class User extends Model<InferAttributes<User>, InferCreationAttributes<User>> {
+  declare id: CreationOptional<number>;
+  declare firstName: string;
+  declare lastName: string;
+  declare username: string;
+  declare password: string;
+  declare refreshToken: CreationOptional<string>;
```
sandje00, line 15: I have rewritten the user model this way because I thought it would help solving the problem described in the first comment. Although it has not helped, this approach to TS Sequelize model definition seems more concise than declaring user interface manually and then implementing it on user model. For that reason, I kept this code and I think we should consider adopting this approach in other model definitions.

sandje00 (commented):
Please, could you read the comments before reviewing. 👇

backend/src/user/user.model.ts
```diff
@@ -94,20 +104,43 @@ class User extends Model implements IUser {
...
+  private _generateAccessToken(): string {
+    const secret = process.env.ACCESS_TOKEN_SECRET || '';
+    const options = {
+      audience: Audience.Scope.Access,
+      expiresIn: process.env.ACCESS_TOKEN_DURATION,
+    };
+    return this._generateToken(secret, options);
+  }
+
+  private _generateRefreshToken(): string {
```
Lnola, line 130: `_generateAccessToken` and `_generateAccessToken` are the same function with different parameters

you could add a tokenType like this

```
type ITokenType = 'ACCESS' | 'REFRESH';
```
and then do this

```
private _generateToken(tokenType: ITokenType) {
    const secret = process.env[`${tokenType}_TOKEN_SECRET`] || '';
    const options = {
      audience: Audience.Scope.Refresh,
      expiresIn: process.env[`${tokenType}_TOKEN_DURATION`],
    };

    const { id, username } = this;
    const payload = { id, username };
    return jwt.sign(payload, secret, options);
  }
```

or store the env data inside an object and index it with the key

```
const tokenData = {
      access: {
        secret: process.env.ACCESS_TOKEN_SECRET,
        duration: process.env.ACCESS_TOKEN_DURATION,
      },
      refresh: {
        secret: process.env.REFRESH_TOKEN_SECRET,
        duration: process.env.REFRESH_TOKEN_DURATION,
      },
    };

const secret = tokenData[tokenType].secret || '';
    const options = {
      audience: Audience.Scope.Refresh,
      expiresIn: tokenData[tokenType].duration,
    };
```

Lnola (changes requested):
The logic is good, we just need to tweak a few implementation details for an approve

Lnola (commented):
Oh and please resolve the conflicts with main

backend/package.json
```diff
@@ -4,7 +4,7 @@
   "description": "",
   "main": "index.js",
   "scripts": {
-    "dev": "nodemon -r tsconfig-paths/register src/app.ts",
+    "dev": "nodemon -r tsconfig-paths/register --files src/app.ts",
```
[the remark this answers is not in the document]
sandje00: Yes, it is for custom interface extensions in `./src/@types` to be loaded and applied.

backend/src/shared/auth/authenticate.ts
```diff
@@ -0,0 +1,29 @@
...
+import { FORBIDDEN_ERROR } from 'shared/constants/errorMessages';
+import { IJwtPayloadDecoded } from './types';
+import { User } from '../database';
+
+export default async function authenticate(req: Request, res: Response, next: NextFunction) {
+  const { accessToken } = req.cookies;
+  if (!accessToken) {
+    return res.status(403).json({ error: FORBIDDEN_ERROR });
+  }
+  try {
```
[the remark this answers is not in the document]
sandje00: I will definitely add a blank line to improve readability (sorry, the mentality from some previous projects). The reason I am not using one line if in `authenticate` and `refresh` particularly is so that the lines don't get too long.

backend/src/@types/express/index.d.ts
```diff
@@ -0,0 +1,9 @@
+import { User } from 'shared/database';
+
+declare global {
+  namespace Express {
+    export interface Request {
+      user: User;
+    }
+  }
+}
```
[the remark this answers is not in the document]
sandje00: I believe this is still going to be necessary in order to be able to attach user to `req` object.

backend/src/shared/auth/refresh.ts
sandje00: Regarding the first comment you made: haven't we exported all of the models in `shared/database/index` in order to be able to use them across the application and use files like `user.model.ts` only inside `user` folder (which applies for other entities, as well)?

Lnola: yes, but if you look inside `user.controller.ts` we used the model directly and we said that it should be changed in the future. I'd say that for now it is ok to have it that way and to change it later.

For the last hour I've been trying to solve this issue and I've found a way to do it. Since we don't have a production I'm fine with this going into `main` but it doesn't have to since I'll be changing it in a few minutes 😄

Lnola: I think we should follow what we've done in user.controller and push it like that (so from the model, not from the index.ts), even though it's not ideal, it is the way we did it and we should be consistent. 

That way we have a consistent base to come to and changes to it are done in a different PR since many side changes in one PR are generally not a good idea - separate issues in their separate pull requests

backend/src/shared/auth/authenticate.ts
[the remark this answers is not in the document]
Lnola: just an FYI, it's fine if it stays like that (one liners I mean), but to note that we've set the `printwidth` in `prettierrc` so you don't have to worry about lines being to long - prettier will automatically put them in new line if they are too long

backend/src/@types/express/index.d.ts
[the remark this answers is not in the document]
Lnola: I'd say this is a bit overkill when you can just do 

```
interface IUserRequest extends Request {
  user: typeof User;
}
```

or

```
type IUserRequest = Request & {
  user: typeof User;
};
```

backend/src/shared/auth/refresh.ts
```diff
@@ -0,0 +1,45 @@
...
+  }
+
+  try {
+    const user = await User.findOne({ where: { refreshToken } });
+    const { id, username, aud } = jwtVerify(refreshToken, process.env.REFRESH_TOKEN_SECRET || '') as IJwtPayloadDecoded;
+
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+    // @ts-ignore
+    if (!user || id !== user.id || username !== user.username || aud !== Audience.Scope.Refresh) {
+      return res.status(403).json({ error: errorMessages.FORBIDDEN_ERROR });
```
Lnola, line 24: think you missed one 😄 
`error` -> `message`

backend/src/config/authTokens.ts
```diff
@@ -0,0 +1,22 @@
+import Audience from 'shared/auth/audience';
```
Lnola, line 1: can you explain why this is in the config folder in src. I would probably put it in `shared/constants` or in `shared/auth`

backend/src/shared/auth/helpers.ts
```diff
@@ -0,0 +1,10 @@
+import { IGeneratedTokens } from 'user/types';
```
Lnola, line 1: nitpick - I'd put this in `shared/helpers/auth` or `shared/utils/auth`

backend/src/shared/auth/authenticate.ts
```diff
@@ -0,0 +1,34 @@
...
+import { User } from '../database';
+
+const authenticate = async (req: Request, res: Response, next: NextFunction) => {
+  const { accessToken } = req.cookies;
+  if (!accessToken) {
+    return res.status(403).json({ message: errorMessages.FORBIDDEN_ERROR });
+  }
+
+  try {
+    const { id, aud } = jwtVerify(accessToken, process.env.ACCESS_TOKEN_SECRET || '') as IJwtPayloadDecoded;
```
abasic, line 15: There was no value in this [types](https://www.npmjs.com/package/@types/jsonwebtoken) package?

backend/src/shared/auth/authenticate.ts
```diff
@@ -0,0 +1,34 @@
...
+    const user = await User.findByPk(id);
+
+    if (!user || aud !== Audience.Scope.Access) {
```
abasic, line 18: Nitpick: the audience can be checked before fetching the user so there is no need to do the database query.

backend/src/shared/auth/authenticate.ts
```diff
@@ -0,0 +1,34 @@
...
+import { NextFunction, Request, Response } from 'express';
+import Audience from './audience';
+import errorMessages from 'shared/constants/errorMessages';
+import { IJwtPayloadDecoded } from './types';
+import { User } from '../database';
+
+const authenticate = async (req: Request, res: Response, next: NextFunction) => {
+  const { accessToken } = req.cookies;
+  if (!accessToken) {
+    return res.status(403).json({ message: errorMessages.FORBIDDEN_ERROR });
```
abasic, line 11: Nitpick: There are [npm packages](https://npms.io/search?q=http+status) that can be used instead of the number for the status code. They are more informative.

Lnola (changes requested):
I think `@types/express/index.d.ts` and using the `user.model.ts` need to be changed or at least commented about more before we can merge with main

backend/src/shared/auth/authenticate.ts
sandje00: I guess this should be implemented as a part of global error handling PR.

backend/src/config/authTokens.ts
sandje00: Because you can reason about this file as configuration for further token usage.

backend/src/@types/express/index.d.ts
```diff
@@ -0,0 +1,9 @@
+import { User } from 'shared/database';
+
+declare global {
+  namespace Express {
+    export interface Request {
+      user: User;
```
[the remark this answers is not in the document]
Lnola: the fix is - put it as `typeof User`

however, there is a problem with that. If you do that you get this error in `authenticate.ts` on `req.user`

`Type 'User' is not assignable to type 'ModelStatic<Model<any, any>>'.`

That is the same error cause by importing user from `database/index.ts` as the other ones

backend/src/config/authTokens.ts
Lnola: makes sense

backend/src/@types/express/index.d.ts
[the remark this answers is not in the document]
Lnola: and leaving a comment here since I'm not getting a response to the other one

I'm against using a declaration file for solving something we created. Declaration files describe the shape of external libraries and help typescript understand what they should look like. Since `User` is not a normal part of Express, but something we add to the existing express `Request` inside our own code it should be an interface or type that only extends the `Request` with the `user` attribute (that is type `typeof User`)

sorry for being a pain 😕

abasic (approved):

Lnola (approved):

PR #20, opened 2022-07-18
Sign in register FE

### LINKED ISSUE #18: Sign in and register FE
- [ ] Login
- [ ] Register

Lnola:
This PR handles the sign in and register forms on the frontend. Along with the forms, there is the initial router setup, minor changes to global styling (that is currently not in use elsewhere!!!!), etc.

The initial router setup was implemented with only `/auth` since not all routes have been defined yet. 
To test - go to `/auth`. You should see the login form first. Submitting still does nothing (should be coming in #15 or another separate PR). The first input `Username` should be autofocused. The entire navigation should be accessible using just the keyboard. 

Pressing the `Click here to register` button should make the current form disappear, the yellow slider to slide from left to right and the Register form to appear. On it, the first input `Username` should be autofocused, as was on the Sign in screen. 

Closes #18


frontend/src/components/auth/UserForm.vue
```diff
@@ -0,0 +1,72 @@
...
+      type: Array,
+      required: true,
+    },
+    buttonLabel: {
+      type: String,
+      default: null,
+    },
+    formSwitch: {
+      type: Object,
+      default() {
```
sandje00, line 40: Since we are not using `this` inside the function, we could rewrite it to an arrow function: `default: () => ({})`.

frontend/src/assets/stylesheets/common/global.css
```diff
@@ -9,9 +9,20 @@ body {
...
+  padding: 0;
+  margin: 0;
 }
 
 #app {
   width: 100%;
   height: 100%;
 }
+
+.hide-element {
```
sandje00, line 26: Nothing major, but the name of this CSS class reminds of an action. Maybe it would be a bit more intuitive to name it something like `hidden-element` or `display-none`?

sandje00 (commented):
Nothing major, just two nitpicks. I love the usage of generic `component` in `UserAuth.vue`! 😄

frontend/src/assets/stylesheets/common/global.css
Lnola: Since I'm not using this and since vue has `v-if` and `v-show` I'd say it's actually better to remove this class entirely

frontend/src/components/auth/UserRegister.vue
```diff
@@ -0,0 +1,26 @@
...
+
+<script>
+import UserForm from './UserForm.vue';
+
+export default {
+  name: 'user-register',
+  setup() {
+    const formInputs = [
+      { label: 'Username', type: 'text' },
+      { label: 'Fist name', type: 'text' },
```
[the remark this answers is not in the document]
Lnola: users register is so through we even ask our users to tell us what they call their fist 👊

abasic (approved):

PR #22, opened 2022-07-19
Tag selection

### LINKED ISSUE #21: Tags

Lnola:
Rebased from main, this PR contains the tag screen which involves a list of possible tags to select. To test, there is the new route - `/tags`. Two new components - `TagSelect` is the main one containing the title, subtitle and actions and `TagList`.

Keep in mind while reviewing that the `TagList` component which is meant to be reused elsewhere, not just in this use-case.

Since these changes need to wait for two other pull requests to be finished I've marked things I can't finish yet with a `// TODO: description...` comment (should light up orange if you're using the [todo highlight](https://marketplace.visualstudio.com/items?itemName=wayou.vscode-todo-highlight) extension)

If tags are not loading it's because of CORS which will be set up in a different PR. Just add an allow CORS extension to your browser ([link](https://chrome.google.com/webstore/detail/allow-cors-access-control/lhobafahddgcelffkeicbaginigeejlf?hl=en) to save time if you don't have it installed)

Closes #21


frontend/src/api/tag.js
```diff
@@ -0,0 +1,14 @@
+import { extractData } from './helpers';
+import request from './config';
+
+const urls = {
+  root: '/tags',
+};
+
+function fetch() {
```
sandje00, line 8: Can we use more descriptive name, e. g. `fetchAll`?

frontend/src/components/tags/TagList.vue
```diff
@@ -0,0 +1,40 @@
...
+  props: {
+    tags: { type: Array, required: true },
+  },
+  components: { ChiButton },
+};
+</script>
+
+<style scoped>
+.tag-list {
+  direction: ltr;
```
sandje00, line 30: Just a question; isn't `ltr` a default value for property `direction`?

frontend/src/components/tags/TagSelect.vue
```diff
@@ -0,0 +1,103 @@
...
+        tags.push({ ...tag, selected: false });
+      });
+    });
+
+    const selectTag = tagId => {
+      const tag = tags.find(tag => tag.id === tagId);
+      tag.selected = !tag.selected;
+    };
+
+    const areAnySelected = () => {
```
sandje00, line 40: Do you think it makes sense to make `areAnySelected` a computed property?

frontend/src/api/tag.js
Lnola: renamed to `fetchAll`

frontend/src/components/tags/TagList.vue
Lnola: Yes, it is. The reason I did this is a CSS "hack" to set the scrollbar left instead of right. I switched the parents default direction to `rtl` to swap the position of the scrollbar to left but then I had to revert the content direction back to the left with this line. Since the parent is the one with the `overflow: scroll` attribute it has the scrollbar and by setting it's direction to the left I set the scrollbar left, but then had to revert the content back so only the scrollbar moves

frontend/src/api/config.js
```diff
@@ -0,0 +1,10 @@
+import axios from 'axios';
+
+const config = {
+  baseURL: import.meta.env.VITE_API_PATH,
```
abasic, line 4: Can baseURL here be relative or does it need to be absolute (like it is defined in `docker-compose.yaml`?

abasic: And can something be done the `vite` way?

frontend/src/components/tags/TagList.vue
abasic: 👌 (Can not react with this one so I am writing a comment).

frontend/src/components/tags/TagSelect.vue
```diff
@@ -0,0 +1,103 @@
...
+export default {
+  name: 'tag-select',
+  setup() {
+    const router = useRouter();
+
+    const tags = reactive([]);
+
+    onMounted(async () => {
+      const tagsArray = await tagApi.fetchAll();
+      tagsArray.forEach(tag => {
```
abasic, line 30: Can `map` be used instead? Or is there an issue with the `reactive` concept?

frontend/src/components/tags/TagSelect.vue
```diff
@@ -0,0 +1,103 @@
+<template>
+  <main class="select-main">
+    <div class="select-title-wrapper">
+      <h1 class="select-title">Select your favorite music genres</h1>
+      <h3 class="select-subtitle">Tell us your favorite music genres so you see the rooms you want to visit first!</h3>
+    </div>
+
+    <tag-list @select-tag="selectTag" class="select-tags" :tags="tags"></tag-list>
+    <chi-button v-if="areAnySelected()" @click="handleSaveSelected" class="select-skip" inline>Continue >></chi-button>
```
abasic, line 9: I believe there is no need for the brackets when using a `computed`.

frontend/src/components/tags/TagSelect.vue
```diff
@@ -0,0 +1,103 @@
...
+}
+
+.select-tags {
+  max-width: 85%;
+  max-height: calc(88px * 2);
+  overflow-y: scroll;
+  margin-bottom: 20px;
+  direction: rtl;
+  padding-left: 20px;
+  margin-left: -20px;
```
abasic, line 96: I haven't run this yet (my bad) but I am wary of negative margines. I assume this is because of the scroll on the left side. Could you please expand as to what this is trying to achieve?

frontend/src/components/tags/TagSelect.vue
sandje00, line 9: It is possible to use a computed property like a variable, so `()` is not necessary.

frontend/src/components/tags/TagSelect.vue
```diff
@@ -0,0 +1,103 @@
+<template>
+  <main class="select-main">
+    <div class="select-title-wrapper">
+      <h1 class="select-title">Select your favorite music genres</h1>
+      <h3 class="select-subtitle">Tell us your favorite music genres so you see the rooms you want to visit first!</h3>
+    </div>
+
+    <tag-list @select-tag="selectTag" class="select-tags" :tags="tags"></tag-list>
+    <chi-button v-if="areAnySelected()" @click="handleSaveSelected" class="select-skip" inline>Continue >></chi-button>
+    <chi-button v-else @click="handleProceed" class="select-skip" inline>Skip for now >></chi-button>
```
abasic, line 10: From the users perspective, what is gained by not allowing them to skip even though they have selected something?

abasic (changes requested):

frontend/src/components/tags/TagSelect.vue
Lnola: Valid point. It was a way of decluttering using just 1 button instead of 2 but UX is more important than UI, I'll update this!

frontend/src/components/tags/TagSelect.vue
Lnola: I was having issues with making the array reactive - couldn't save the fetched data directly into it, couldn't use `concat`, only push seemed to work. After I got this working I ended up changing a few things so I'll look into if this is possible to do now.

frontend/src/api/config.js
abasic: The answers to these two questions are different so reading them again I see they are confusing. Sorry about that.

The URL is absolute in the `docker-compose.yaml` (http://localhost:3001). The browser by default sends the payload to the domain of the webpage that is hosting the site. In this case, it would be http://localhost:3000.

Starting from the second question, `vite` has a [proxy](https://vitejs.dev/config/server-options.html#server-proxy) option which I believe should be leveraged for this. In development, the frontend is hosted by a vite server that is different from the backend server (and thus a different URL) but (usually) in production they should be behind the same URL. So having the `axios` configuration reference the URL of the backend directly does not seem like a clean solution. The development can mimic a typical production environment (frontend and backend behind the same URL) and avoid coupling the frontend code with the configuration.

After answering the second question the first becomes obsolete in my eyes. At least from the perspective of this project. But the answer would be that it can be relative, and it probably should. It will come into play if you put some of your prefix your routes with `/api/`. Which is a pattern some follow. Hope this clears it up. Sorry again for the confusing questions.

frontend/src/components/tags/TagSelect.vue
abasic: Fair enough 😄 . Just a note since you are playing around with it: the scroll behavior on Mac OS can be changed to work like Windows (always visible). You can change it [here](https://www.lifewire.com/change-how-macs-scroll-bars-work-2260834). If you have not seen it, it might be useful for playing around with this and seeing how it behaves.

frontend/src/api/config.js
Lnola: This makes so much sense! I'm glad you explained it since I haven't had much experience with something like this. I'll play around with it a bit now. Never thought about this before, will do for this project and in the future! 🚀

backend/src/app.ts
```diff
@@ -13,10 +14,10 @@ const port = process.env.PORT || 3001;
...
 app.use(express.json());
 
-app.use(message.path, message.router);
-app.use(room.path, room.router);
-app.use(tag.path, tag.router);
-app.use(user.path, user.router);
+app.use(createEndpoint(message.path), message.router);
+app.use(createEndpoint(room.path), room.router);
+app.use(createEndpoint(tag.path), tag.router);
+app.use(createEndpoint(user.path), user.router);
```
abasic, line 20: Nitpick: Another way to do this is to create a new parent router and use it instead of `app` so that all these subroutes are set up on it. And then set that parent router to be on `/api`. In code terms something like this:

```js
import { Router } from 'express';

const router = Router();

router.use(message.path, message.router);
router.use(room.path, room.router);
router.use(tag.path, tag.router);
router.use(user.path, user.router);

app.use('/api', router);
```
I believe this is a cleaner approach that leverages how `express` handles routes.

frontend/src/api/tag.js
```diff
@@ -0,0 +1,14 @@
+import { extractData } from './helpers';
+import request from './config';
+
+const urls = {
+  root: '/api/tags',
```
abasic, line 5: Every new URL set up this way will start with `/api`. Is there a way to make it so their URLs need to only reference the part of the URL that is unique to them? Hint: this is connected to the `axios` `baseURL` question from earlier 😄

frontend/src/components/tags/TagList.vue
```diff
@@ -0,0 +1,40 @@
...
+        class="tag-button"
+      >
+        {{ label }}
+      </chi-button>
+    </div>
+  </div>
+</template>
+
+<script>
+import ChiButton from '../common/ChiButton.vue';
```
abasic, line 18: Hint: such common components can be [registered globally](https://vuejs.org/guide/components/registration.html#global-registration) so they do not need to be imported into each component that uses them 🤓

frontend/src/components/tags/TagSelect.vue
```diff
@@ -0,0 +1,108 @@
...
+
+export default {
+  name: 'tag-select',
+  setup() {
+    const router = useRouter();
+
+    const tags = reactive([]);
+
+    onMounted(async () => {
+      tags.push(...(await tagApi.fetchAll()));
```
abasic, line 29: I'll look into if this can be done that does not rely on push.

frontend/src/components/tags/TagSelect.vue
```diff
@@ -0,0 +1,108 @@
...
+      tags.map(tag => (tag.selected = false));
```
abasic, line 30: The intention for `Array.map` is to be used to create a new array that is based on an existing one. Since here it is only used to loop through the existing array, `forEach` is more appropriate.

abasic (changes requested):

frontend/src/components/tags/TagSelect.vue
abasic: I played around with it and it seems this can not be done when making an array reactive with the `reactive` method. The `ref` method needs to be used. Their [documentation](https://vuejs.org/guide/essentials/list.html#array-change-detection) also mentions it.

backend/src/app.ts
Lnola: much cleaner, love it 😍

frontend/src/components/tags/TagSelect.vue
Lnola: I can do something like this when using `ref`

```
const tagsArray = await tagApi.fetchAll();
tags.value = tagsArray.map(tag => ({ ...tag, selected: false }));
```

it can even be a one-liner but I think something like this looks cleaner

abasic (dismissed):

sandje00 (approved):
🚀

PR #23, opened 2022-07-19
Move @type dependencies to devDependencies

Lnola:
As the title says, this PR moves the `@type` dependencies from the backend to the dev dependencies list (where they belong). When installing the future changes make sure to delete your local `backend/node_modules` before running `docker compose up`!!!!!


sandje00 (approved):
✅

abasic (approved):
@Lnola Does your comment about deleting `node_modules` relate to only future backend changes or frontend as well? Or is the future this PR? 😄

Lnola:
> @Lnola Does your comment about deleting `node_modules` relate to only future backend changes or frontend as well? Or is the future this PR? 😄

only backend `node_modules` need to be reinstalled since they were the only ones changed. The future is this PR and the `main` branch after it is merged into it. What i meant by that comment is - if anyone pulled the repo before this change, they need to reinstall the `node_modules` 😄

PR #24, opened 2022-07-21
Bump vite to v3 and configure proxy

Lnola:
Bumped vite to v3 to fix proxy problems
Updated config - configured the correct proxy url and vite dev server ports
Removed baseURL from axios config

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -19,7 +19,7 @@
     "@babel/eslint-parser": "^7.18.2",
     "@babel/eslint-plugin": "^7.17.7",
     "@extensionengine/eslint-config": "^3.0.1",
-    "@vitejs/plugin-vue": "^2.3.3",
+    "@vitejs/plugin-vue": "3",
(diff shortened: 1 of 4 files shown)


frontend/package.json
```diff
@@ -19,7 +19,7 @@
     "@babel/eslint-parser": "^7.18.2",
     "@babel/eslint-plugin": "^7.17.7",
     "@extensionengine/eslint-config": "^3.0.1",
-    "@vitejs/plugin-vue": "^2.3.3",
+    "@vitejs/plugin-vue": "3",
```
abasic, line 22: All the other ones are semver so please change this and the other one to be semver.

abasic (changes requested):
Just the minor thing with the versions in `package.json`.

frontend/package.json
Lnola: Nice catch, updating now!

abasic (approved):

PR #26, opened 2022-07-25
Global Error Handling

sandje00:
THIS PR:

- Add global error handler
- Add `HttpError` class
- Rewrite the existing code using global error handling


backend/src/app.ts
```diff
@@ -24,3 +25,5 @@ app.use('/api', router);
 app.listen(port, function () {
   console.log(`App is listening on port ${port} !`);
 });
+
+app.use(errorHandler);
```
Lnola, line 29: why is this error middleware after app.listen but the others are before it. What is the reason and does it make a difference? 🤔

backend/src/shared/auth/authenticate.ts
```diff
@@ -2,24 +2,26 @@ import { JsonWebTokenError, JwtPayload, verify as jwtVerify, TokenExpiredError }
...
+import { FORBIDDEN } from 'http-status';
+import HttpError from 'shared/error/httpError';
 import { IAuthRequest } from './types';
 import { User } from 'shared/database';
 
 const authenticate = async (req: IAuthRequest, res: Response, next: NextFunction) => {
   const { accessToken } = req.cookies;
   if (!accessToken) {
-    return res.status(403).json({ message: errorMessages.FORBIDDEN_ERROR });
+    return next(new HttpError(FORBIDDEN, errorMessages.FORBIDDEN_ERROR));
```
Lnola, line 13: ❤️

backend/src/shared/error/errorHandler.ts
```diff
@@ -0,0 +1,12 @@
+import { NextFunction, Response } from 'express';
+import { IAuthRequest } from 'shared/auth/types';
+import { IHttpError } from './types';
+import { INTERNAL_SERVER_ERROR } from 'http-status';
+
+const errorHandler = (err: IHttpError | Error, _req: IAuthRequest, res: Response, _next: NextFunction) => {
+  return 'status' in err
+    ? res.status(err.status).json({ message: err.message })
+    : res.status(INTERNAL_SERVER_ERROR).json({ message: 'Something went wrong' });
+};
```
abasic, line 10: Would something like this be more readable? I just assumed some TypeScript stuff 😄 
```js
const DEFAULT_ERROR: IHttpError = {
  status: INTERNAL_SERVER_ERROR,
  message: 'Something went wrong'
};

const errorHandler = (err: IHttpError | Error, _req: IAuthRequest, res: Response, _next: NextFunction) => {
  const error = 'status' in err ? err : DEFAULT_ERROR;
  return res.status(error.status).json({ message: error.message });
};
```

backend/src/shared/error/errorHandler.ts
```diff
@@ -0,0 +1,12 @@
+import { NextFunction, Response } from 'express';
+import { IAuthRequest } from 'shared/auth/types';
+import { IHttpError } from './types';
+import { INTERNAL_SERVER_ERROR } from 'http-status';
+
+const errorHandler = (err: IHttpError | Error, _req: IAuthRequest, res: Response, _next: NextFunction) => {
```
Lnola, line 6: I'm getting a TS error for `_next`, guessing since it's last and isn't used it is viewed as a lint error. Since we have `husky` you should've gotten this exact warning about this (and if we keep this there will be an error about this every time we commit).

```
/Users/eedev/projects/chichet/backend/src/shared/error/errorHandler.ts
  6:83  warning  '_next' is defined but never used  @typescript-eslint/no-unused-vars

✖ 1 problem (0 errors, 1 warning)
```

I got this when I tried to commit just to test if it works. If it's not working for you let me know so we can fix it!

backend/src/user/user.controller.ts
```diff
@@ -10,32 +12,32 @@ const getAll = async (req: Request, res: Response) => {
...
 
-    return res.status(201).send();
+    return res.status(CREATED).send();
   } catch (err) {
     if (err instanceof UniqueConstraintError) {
-      return res.status(409).json({ message: errorMessages.REGISTER_ERROR });
+      return next(new HttpError(CONFLICT, errorMessages.REGISTER_ERROR));
     }
-    return res.status(500).end();
+    return next(new HttpError(INTERNAL_SERVER_ERROR));
```
abasic, line 40: Will sending the status here mean that the JSON returned to the frontend will not have a message?

abasic (commented):
Two very minor comments 👍

backend/src/shared/error/errorHandler.ts
```diff
@@ -0,0 +1,12 @@
...
+  return 'status' in err
```
Lnola, line 7: change the `err: IHttpError | Error` to `err: IHttpError` and then this can be `err.status` instead of this workaround

Lnola (changes requested):

backend/src/user/user.controller.ts
sandje00: I meant to send built-in error in this line and leave it to `errorHandler` to add status code and error message, but I forgot.

backend/src/shared/error/errorHandler.ts
sandje00: I am aware of this Husky warning. Removing this next function is not an option because the global error handling is not going to work properly. The only solution that I see is putting `eslint-disable-next-line` comment.

backend/src/app.ts
sandje00: Actually, I checked it and the only important thing is that global error handler is used after all of the other route handlers, so it can catch all errors. Should I move it above `app.listen()`?

Lnola: I couldn't find any reasons why it should be before of after `app.listen`, it can stay as is!

backend/src/shared/error/errorHandler.ts
abasic: Then let's add the eslint comment so it skips the next line.

Lnola (dismissed):

Lnola (approved):

PR #31, opened 2022-07-26
Persist user on login (protected routes)

### LINKED ISSUE #32: Protected routes/persist user

Lnola:
This PR adds the user to the frontend persistent storage (`localstorage`) and pinia store. By modifying `signIn` and `register` on the backend to return the authenticated/created user the frontend is able to store the user in `localstorage` and access, set... it through pinia store. 

The pinia store uses `Composition API` syntax and is connected to `localstorage` with `@vueuse/core` `useStorage` method.

How it works and what to test
1.  a user that is not signed in cannot access any route except `/auth`
2. a signed in user cannot access `/auth` route
3. the user information should be saved in `localstorage` as an object (empty if not signed in)

Closes #32

### DIFF
--- a/frontend/src/components/auth/UserRegister.vue
+++ b/frontend/src/components/auth/UserRegister.vue
@@ -12,6 +12,7 @@ import { setErrorToLastForm, validateAuthForm } from '@/utils/validation';
 import { auth as authApi } from '@/api';
 import { reactive } from 'vue';
 import { registerFormLabels } from './constants';
+import { useAuthStore } from '@/store/authStore';
 import UserForm from './UserForm.vue';
 import { useRouter } from 'vue-router';
 
@@ -19,6 +20,10 @@ export default {
   name: 'user-register',
   setup() {
     const router = useRouter();
+    const userStore = useAuthStore();
+
+    console.log(userStore.getUser);
+
     const formInputs = reactive(registerFormLabels.formInputs);
 
     const submit = async () => {
@@ -30,7 +35,8 @@ export default {
       const password = formInputs.password.value;
 
       try {
-        await authApi.register({ username, firstName, lastName, password });
+        const user = await authApi.register({ username, firstName, lastName, password });
+        userStore.setUser(user);
         router.push({ name: 'Tags' });
       } catch (err) {
         setErrorToLastForm(formInputs, err.response.data.message);
--- a/frontend/src/components/auth/UserRegister.vue
+++ b/frontend/src/components/auth/UserRegister.vue
@@ -22,8 +22,6 @@ export default {
     const router = useRouter();
     const userStore = useAuthStore();
 
-    console.log(userStore.getUser);
-
     const formInputs = reactive(registerFormLabels.formInputs);
 
     const submit = async () => {
--- a/frontend/src/components/auth/UserSignIn.vue
+++ b/frontend/src/components/auth/UserSignIn.vue
@@ -12,7 +12,7 @@ import { setErrorToLastForm, validateAuthForm } from '@/utils/validation';
 import { auth as authApi } from '@/api';
 import { reactive } from 'vue';
 import { signInFormLabels } from './constants';
-import { useAuthStore } from '../../store/authStore';
+import { useAuthStore } from '@/store/authStore';
 import UserForm from './UserForm.vue';
 import { useRouter } from 'vue-router';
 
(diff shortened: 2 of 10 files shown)


frontend/src/components/auth/UserRegister.vue
```diff
@@ -12,13 +12,18 @@ import { setErrorToLastForm, validateAuthForm } from '@/utils/validation';
...
 import UserForm from './UserForm.vue';
 import { useRouter } from 'vue-router';
 
 export default {
   name: 'user-register',
   setup() {
     const router = useRouter();
+    const userStore = useAuthStore();
+
+    console.log(userStore.getUser);
```
abasic, line 25: 👀

frontend/src/components/auth/UserSignIn.vue
```diff
@@ -12,13 +12,16 @@ import { setErrorToLastForm, validateAuthForm } from '@/utils/validation';
 import { auth as authApi } from '@/api';
 import { reactive } from 'vue';
 import { signInFormLabels } from './constants';
+import { useAuthStore } from '../../store/authStore';
```
abasic, line 15: Sometimes `@` is used, and sometimes it is not. Please be consistent.

frontend/src/router/index.js
```diff
@@ -26,4 +27,16 @@ const router = createRouter({
   routes,
 });
 
+router.beforeEach(async (to, from) => {
```
abasic, line 30: I believe `async` is not needed here.

frontend/src/router/index.js
```diff
@@ -26,4 +27,16 @@ const router = createRouter({
...
+  const user = useAuthStore().getUser;
+
+  if (!user.id && to.name !== 'Auth') {
+    return { name: 'Auth' };
+  }
+
+  if (!!user.id && to.name === 'Auth') {
+    return { name: 'Home' };
+  }
```
abasic, line 39: I believe this code can be rewritten so it is more easily understood and extended down the line. For example:
```js
const isLoggedIn = !!user.id;
const isAuthRoute = to.name === 'Auth';

if (!isLoggedIn && !isAuthRoute) return { name:  'Auth'  };
if (isLoggedIn && isAuthRoute) return { name: 'Home'  };
```
or
```js
const isLoggedIn = !!user.id;
const isAuthRoute = to.name === 'Auth';

if (isLoggedIn) {
  if (isAuthRoute) return { name: 'Home' };
} else {
  if (!isAuthRoute) return { name: 'Auth' };
}
```
The general idea is to extract some parts of the logic into variables so it is easier to keep track of what is happening and what is being used to decide where to redirect.

abasic (changes requested):

frontend/src/components/auth/UserRegister.vue
Lnola: hahahhaha, you saw NOTHING!!!

frontend/src/components/auth/UserSignIn.vue
Lnola: I try to be consistent with using @, this was an auto import that managed to slip through! I'll be more attentive to it next time!

frontend/src/router/index.js
Lnola: Yep, forgot to remove it when I removed the authenticate call (old logic)

frontend/src/router/index.js
Lnola: I agree (I went with the first option)

frontend/src/store/authStore.js
```diff
@@ -0,0 +1,17 @@
+import { computed, ref } from 'vue';
+import { defineStore } from 'pinia';
+import { useStorage } from '@vueuse/core';
+
+export const useAuthStore = defineStore('auth', () => {
+  const user = ref(useStorage('user', {}));
```
sandje00, line 6: Nice! 🚀

sandje00 (dismissed):
Everything looks good! 😃

frontend/src/router/index.js
```diff
@@ -26,4 +27,17 @@ const router = createRouter({
   routes,
 });
 
+router.beforeEach((to, from) => {
+  const isLoggedIn = !!useAuthStore().getUser.id;
```
abasic, line 31: This could have been a [getter](https://pinia.vuejs.org/core-concepts/getters.html) in pinia. Now it is only used here but it might come into play later on.

abasic (dismissed):
One small note for future sake. :D

frontend/src/router/index.js
Lnola: I like that idea, `id` should be used a lot later

abasic (approved):

PR #35, opened 2022-07-26
Sign out

### LINKED ISSUE #33: Sign out

Lnola:
Create user sign out - includes endpoint for sign out which deletes the tokens, `authStore` implementation of the sign out action and a live example on route `/` with the `Sign out` button

Closes #33

### DIFF
--- a/backend/src/user/user.controller.ts
+++ b/backend/src/user/user.controller.ts
@@ -43,4 +43,11 @@ const register = async ({ body }: IRegisterRequest, res: Response, next: NextFun
   }
 };
 
-export { getAll, signIn, register };
+const signOut = (req: Request, res: Response) => {
+  res.clearCookie('accessToken');
+  res.clearCookie('refreshToken');
+
+  res.status(OK).send();
+};
+
+export { getAll, signIn, register, signOut };
--- a/backend/src/user/user.controller.ts
+++ b/backend/src/user/user.controller.ts
@@ -1,9 +1,9 @@
+import { clearAuthCookies, setAuthCookies } from 'shared/helpers/auth';
 import { CONFLICT, CREATED, OK, UNAUTHORIZED } from 'http-status';
 import { IRegisterRequest, ISignInRequest } from './types';
 import { NextFunction, Request, Response } from 'express';
 import errorMessages from 'shared/constants/errorMessages';
 import HttpError from 'shared/error/httpError';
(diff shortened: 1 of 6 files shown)


frontend/src/store/authStore.js
```diff
@@ -13,5 +15,11 @@ export const useAuthStore = defineStore('auth', () => {
     user.value = userToSet;
   };
 
-  return { user, getUser, setUser };
+  const signOut = async router => {
```
abasic, line 18: Does the router need to be passed as a prop or could it have been imported at the top?

frontend/src/components/HomeTemp.vue
```diff
@@ -1,9 +1,20 @@
...
+    <button @click="authStore.signOut(router)">Sign out</button>
+  </div>
 </template>
 
 <script>
+import { useAuthStore } from '../store/authStore';
+import { useRouter } from 'vue-router';
+
 export default {
   name: 'home-temp',
```
abasic, line 12: Is this still a good name for it? 😄

frontend/src/store/authStore.js
Lnola: It does. I first tried by importing directly in the `authStore.js` and it returned `inject() can only be used inside setup() or functional components`. That's why it had to be a prop

frontend/src/components/HomeTemp.vue
Lnola: I would say it is ok since soon I'll be working on replacing this entire component, this is just for testing purposes and for the sake of routing working with route names instead of paths

backend/src/user/user.controller.ts
```diff
@@ -43,4 +43,11 @@ const register = async ({ body }: IRegisterRequest, res: Response, next: NextFun
   }
 };
 
-export { getAll, signIn, register };
+const signOut = (req: Request, res: Response) => {
+  res.clearCookie('accessToken');
+  res.clearCookie('refreshToken');
```
sandje00, line 48: Not a strict request to change, rather a proposal; should we extract this cookie clearing logic into separate function in a similar fashion to `setAuthCookies()` in `backend/src/shared/helpers/auth` so we can reuse it in case when both access and refresh token have expired?

Lnola: I like that idea, moving it!

abasic (approved):

PR #37, opened 2022-07-27
Protect endpoints and error interceptor

### LINKED ISSUE #36: Protect endpoints and interceptor

Lnola:
Protect current routes - `tags` with `authenticate` and `refresh` middleware and add axios interceptor to remove the user if the response is `FORBIDDEN`

To test, set the expiry of refresh token to about `30s`. Register/sign in on `/auth` and go to `tags` after. You should see the tags displayed correctly. Wait for the expiry time you set to pass and then refresh the page. You should be redirected to `auth` and your user should be removed from `localstoreage`

Closes #36

### DIFF
--- a/frontend/src/api/request.js
+++ b/frontend/src/api/request.js
@@ -1,4 +1,8 @@
 import axios from 'axios';
+import { StatusCodes } from 'http-status-codes';
+import { useAuthStore } from '@/store/authStore';
+
+const { FORBIDDEN } = StatusCodes;
 
 const config = {
   baseURL: '/api',
@@ -6,5 +10,19 @@ const config = {
 };
 
 const client = axios.create(config);
+
+const isAuthError = err => [FORBIDDEN].includes(err.response.status);
+
+client.interceptors.response.use(
+  res => res,
+  err => {
+    if (isAuthError(err)) {
+      useAuthStore().removeUser();
+      return window.location.reload();
+    }
+
+    throw err;
+  },
+);
 
 export default client;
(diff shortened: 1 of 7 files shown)


frontend/src/api/request.js
```diff
@@ -7,4 +11,18 @@ const config = {
...
+
+client.interceptors.response.use(
+  res => res,
+  err => {
+    if (isAuthError(err)) {
+      useAuthStore().removeUser();
+      return window.location.reload();
+    }
+
+    throw err;
```
sandje00, line 24: Could we use this interceptor to simplify access to non-auth error messages?

Lnola: Yes! We could just add something like `const isNotFoundError = err => [NOT_FOUND].includes(err.response.status);
` and add the handler what we want to do in case of not found error

sandje00: What I initially had in mind was instead of throwing entire `err` object returning only parts of it necessary for error handling inside component (e. g. returning `err.response` to have easier access to `status` and `data.message`). That way we could simplify access to necessary info in `catch` block and ensure to keep the error handling up to implementation of a component.

As of errors that are handled the same way throughout entire application, like auth error, should be handled in this interceptor, for sure. 👍

Lnola: That looks good to me but since we've already been using errors without that some other code, not related to this PR will also need to be refactored. For that reason I'm adding a new task for this instead of doing it in this PR since it is not related to this issue

abasic (approved):

PR #40, opened 2022-07-27
User Profile

### LINKED ISSUE #34: User profile

sandje00:
THIS PR:

- Add user profile container
- Add user profile icon
- Add user profile info (adapt `UserForm.vue` for user data display without edit)
- Display tags in user profile (adapt `TagsList.vue` style for display in user profile)
- Add `fullName` virtual field in `user.model`
- Add `useToggle` composable for boolean value toggling

Notes:

- Button `Toggle User Profile` on page `Home` is just for testing user profile: it should be removed as a part of PR that adds homepage UI.
- Tag fetch and select logic in user profile is just for testing: it will be removed as a part of PR that abstracts tag logic into a composable (marked with a comment in code).

Related issues:
Closes #34


backend/src/user/user.model.ts
```diff
@@ -31,6 +32,13 @@ class User extends Model implements IUser {
         allowNull: true,
       },
 
+      fullName: {
+        type: VIRTUAL,
+        get(this: IUser): string {
```
Lnola, line 37: I don't believe the return type for this function is needed

frontend/src/assets/stylesheets/common/all.css
```diff
@@ -1,3 +1,4 @@
 @import 'colors.css';
 @import 'global.css';
+@import 'global.css';
```
abasic, line 3: This is already imported above.

frontend/src/composables/useToggle.js
```diff
@@ -0,0 +1,9 @@
+import { ref } from 'vue';
+
+export default function useToggle() {
+  const isTrue = ref(false);
```
abasic, line 4: Nitpick: `isToggled` or `isSet` might be better names because `isTrue` does not seem as specific as it needs to be.

frontend/src/components/userProfile/index.vue
```diff
@@ -0,0 +1,122 @@
+<template>
+  <div :class="{ active }" class="user-profile">
+    <div class="user-profile-container">
+      <div class="user-profile-wrapper">
+        <img src="@/assets/img/profile.svg" alt="profile" class="user-profile-avatar" />
+        <user-form :inputs="formInputs" class="user-profile-form" disabled />
```
abasic, line 6: Nitpick: the design does not really reflect that the form is disabled. So, if this is not temporary (set up for another PR), I would advise revisiting this design and making a disabled `user-form` visually different from no active one.

frontend/src/components/userProfile/index.vue
```diff
@@ -0,0 +1,122 @@
...
+    <div @click="$emit('close-profile')" class="user-profile-overlay"></div>
+  </div>
+</template>
+
+<script>
+import { onMounted, reactive, ref } from 'vue';
+import { tag as tagApi } from '@/api';
+import TagList from '../tags/TagList.vue';
+import { useAuthStore } from '@/store/authStore';
+import UserForm from '../auth/UserForm.vue';
```
abasic, line 19: Nitpick: relative paths are used. In #44 that will no longer be allowed. So please update these paths.

frontend/src/components/tags/TagList.vue
```diff
@@ -1,11 +1,12 @@
 <template>
   <div>
-    <div class="tag-list">
+    <div :class="{ centred }" class="tag-list">
       <chi-button
         v-for="{ id, label, selected } in tags"
         :key="id"
         @click="$emit('select-tag', id)"
         :secondary="!selected"
+        :class="{ small }"
```
abasic, line 9: Nitpick: maybe make this an attribute of the `chi-button` component. So that it can dictate how its small style will look like. And that it can be set to be small in other components as well.

backend/src/user/user.controller.ts
```diff
@@ -17,8 +17,8 @@ const signIn = async ({ body: { username, password } }: ISignInRequest, res: Res
   const tokens = await user.generateTokens();
   setAuthCookies(tokens, res);
 
-  const { id, firstName, lastName } = user;
-  return res.status(OK).json({ id, username, firstName, lastName });
+  const { id, firstName, lastName, fullName } = user;
+  return res.status(OK).json({ id, username, firstName, lastName, fullName });
```
abasic, line 21: Hint: by making another virtual field or a getter on the user class and moving all of this information to it, it would simplify any code that requires only this specific subset of the user information. For example:
```js
// user.model.ts
...
profile: {
  type: VIRTUAL,
  get(this: IUser): string {
    return {
      id: this.id,
      username: this.username,
      firstName: this.firstName,
      lastName: this.lastName,
      fullName: this.fullName
    };
  },
},
...

// or

get profile() {
  return {
    id: this.id,
    username: this.username,
    firstName: this.firstName,
    lastName: this.lastName,
    fullName: this.fullName
  };
}

// user.controller.ts
return res.status(OK).json(user.profile);
```
By doing something like this, only that getter or virtual field needs to be updated if new info needs to be added or removed.

abasic (changes requested):
Mostly nitpicks 👍

frontend/src/components/userProfile/index.vue
```diff
@@ -0,0 +1,122 @@
...
+      tag.selected = !tag.selected;
+    };
+
+    return { formInputs, tags, selectTag };
+  },
+  components: { TagList, UserForm },
+};
+</script>
+
+<style scoped>
```
Lnola, line 52: love the transition of the user profile flying in, would love to see it fly away (doesn't need to be added now, we can focus on animations later if we want to do them)

frontend/src/components/userProfile/index.vue
```diff
@@ -0,0 +1,122 @@
...
+  height: 100vh;
+  visibility: hidden;
+}
+
+.user-profile-container {
+  z-index: var(--z-user-profile-container);
+  background-color: var(--color-secondary);
+  border-radius: 0 0 120px 0;
+  width: 50vw;
+  height: 100vh;
```
Lnola, line 68: these two probably could be `%`. I'd avoid using viewport units and especially `vw` (because of the scrollbar issue). The parent can have them, fine, (though I think the parent should have `%` also) but this should be `%` if it can be

frontend/src/components/userProfile/index.vue
```diff
@@ -0,0 +1,122 @@
...
+  position: fixed;
```
Lnola, line 69: this doesn't need to be `fixed` if the parent already is

frontend/src/components/userProfile/index.vue
```diff
@@ -0,0 +1,122 @@
...
+  top: 0;
+  width: 100vw;
+  height: 100vh;
+  visibility: hidden;
+}
+
+.user-profile-container {
+  z-index: var(--z-user-profile-container);
+  background-color: var(--color-secondary);
+  border-radius: 0 0 120px 0;
```
Lnola, line 66: `border-bottom-left-radius: 120px`

frontend/src/components/userProfile/index.vue
Lnola: the position also doesn't need to be set. If you remove these 3 lines (position, top, left) everything still works the same

frontend/src/components/userProfile/index.vue
```diff
@@ -0,0 +1,122 @@
...
+  width: 50vw;
+  height: 100vh;
+  position: fixed;
+  top: 0;
+  left: -100%;
+  transition: 0.6s ease;
+  transition-property: left;
```
Lnola, line 73: shorthand would be: `transition: margin-left 0.6 ease`

Lnola (changes requested):

frontend/src/components/userProfile/index.vue
Lnola: @sandje00 you can use the `--color-primary-subdued` for input coloir and change the text color too to make it look disabled

frontend/src/assets/stylesheets/common/measures.css
```diff
@@ -0,0 +1,5 @@
+:root {
+  --z-user-profile: 10;
+  --z-user-profile-overlay: 20;
+  --z-user-profile-container: 30;
```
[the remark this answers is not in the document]
sandje00: In my view it is useful to have z indices as variables in one place because in case of adding new element that also has z index it is easier to determine such value of its z index that has no collision with existing elements.

frontend/src/components/tags/TagList.vue
```diff
@@ -1,11 +1,12 @@
 <template>
   <div>
-    <div class="tag-list">
+    <div :class="{ centred }" class="tag-list">
```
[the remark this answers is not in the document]
sandje00: I appreciate your sense for details; I agree that we should stick to the convention. I will fix it, but after my tea time. 🍵

abasic:
@sandje00 please resolve the conflicts as well.

frontend/src/components/userProfile/index.vue
sandje00: Maybe we could add entire animation as a part of separate PR.

frontend/src/components/profile/ProfileSidebar.vue
```diff
@@ -0,0 +1,121 @@
+<template>
+  <aside :class="{ active }" class="user-profile">
+    <div class="user-profile-container">
+      <div class="user-profile-wrapper">
+        <img src="@/assets/img/profile.svg" alt="profile" class="user-profile-avatar" />
+        <user-form :inputs="formInputs" class="user-profile-form" disabled />
```
Lnola, line 6: Can we add some kind of `v-if` to not show the empty `Full name` (since it is optional in user profile creation)

frontend/src/components/profile/ProfileSidebar.vue
```diff
@@ -0,0 +1,121 @@
...
+import UserForm from '@/components/auth/UserForm.vue';
+import { userProfileFormLabels } from './constants';
+
+export default {
+  name: 'user-profile',
+  props: {
+    active: { type: Boolean, default: false },
+  },
+  setup() {
+    // Replace with composables
```
Lnola, line 28: can you replace this with `// TODO: Replace with composables`
If you use this [extension](https://marketplace.visualstudio.com/items?itemName=wayou.vscode-todo-highlight) the text lights up orange to catch it easier 🍊

frontend/src/components/profile/ProfileSidebar.vue
```diff
@@ -0,0 +1,121 @@
...
+</script>
+
+<style scoped>
+.user-profile {
+  z-index: var(--z-user-profile);
+  position: fixed;
+  left: 0;
+  top: 0;
+  width: 100vw;
+  height: 100vh;
```
Lnola, line 59: try this instead of setting width and height

```
top: 0;
right: 0;
bottom: 0;
left: 0;
```
or even better, the shorthand
```
inset: 0;
```

just an FYI, `inset` is not supported on every browser so check [this](https://caniuse.com/?search=inset) out before deciding to use it

frontend/src/components/profile/ProfileSidebar.vue
```diff
@@ -0,0 +1,121 @@
...
+  margin-bottom: 10px;
+  direction: rtl;
+  padding-left: 10px;
+  margin-left: -10px;
+}
+
+.user-profile-overlay {
+  z-index: var(--z-user-profile-overlay);
+  width: 100vw;
+  height: 100vh;
```
Lnola, line 108: this can be `%` instead of `vw` and `vh`. Since it is preferred in this scenario I'd change it to be `%`

frontend/src/components/profile/ProfileSidebar.vue
```diff
@@ -0,0 +1,121 @@
...
+  background-color: var(--color-overlay);
+  visibility: hidden;
+}
+
+.user-profile.active,
+.active .user-profile-overlay {
+  visibility: visible;
+}
+
+.active .user-profile-container {
```
Lnola, line 118: nitpick: can this be closer to the original class that has `left: -100%` just fore readability

frontend/src/components/profile/ProfileSidebar.vue
```diff
@@ -0,0 +1,121 @@
...
+.user-profile-overlay {
+  z-index: var(--z-user-profile-overlay);
+  width: 100vw;
+  height: 100vh;
+  background-color: var(--color-overlay);
+  visibility: hidden;
+}
+
+.user-profile.active,
+.active .user-profile-overlay {
```
Lnola, line 114: check this out `.active > .user-profile-overlay`

frontend/src/components/profile/ProfileSidebar.vue
Lnola, line 118: same as the previous comment - `.active > .user-profile-container`

frontend/src/components/profile/constants.js
```diff
@@ -0,0 +1,8 @@
+const userProfileFormLabels = {
+  formInputs: {
+    username: { label: 'Username', type: 'text', value: '', error: '', required: true },
```
Lnola, line 3: just a question, would it work without the `type`, `error` and other unused attributes. If it cannot, then could we make it work by changing the original component?

frontend/src/components/tags/TagList.vue
```diff
@@ -1,12 +1,14 @@
...
   <div>
-    <div class="tag-list">
+    <div :class="{ centered }" class="tag-list">
       <chi-button
         v-for="{ id, label, selected } in tags"
         :key="id"
         @click="$emit('select-tag', id)"
         :secondary="!selected"
-        class="tag-button"
+        :large="!small"
```
Lnola, line 9: could we make it `large` by default so we don't have to do this?

Lnola (changes requested):
Minor changes needed and it will be good to go!

frontend/src/components/common/ChiButton.vue
```diff
@@ -44,4 +56,22 @@ export default {
...
+  border-radius: 100px;
+  padding: 20px 36px;
+}
+
+.chi-button-small {
+  margin: 0 18px 18px 0;
+  padding: 10px 18px;
+  border: 2px solid var(--color-primary);
+  font-size: var(--font-size-small);
+}
```
abasic, line 76: Nitpick: personally I avoid setting margins in common components or in components that I know are intended to be used in different contexts. The reasoning is that the margin property defines how the component interacts with its surroundings (other components around it). So `margin` being defined in them can mess up the flow when being used in a new context. Usually what happens is that then these get overridden by styling outside of the component.

If you choose to do something about this at this point, my recommendation is to have specific classes that apply just the margins. And these classes would not be tied to buttons but could be used on anything.

frontend/src/components/profile/ProfileSidebar.vue
abasic: Could you elaborate why it is prefered in this scenario?

frontend/src/components/tags/TagList.vue
abasic: Or maybe introduce a `size` attribute that consolidates the two.

abasic (commented):
Just minor comments that can be ignored.

frontend/src/components/tags/TagList.vue
sandje00: In that case it would impact "Register" and "SIgn in" buttons and I guess we don't want that.

frontend/src/components/common/ChiButton.vue
sandje00: This is going to be handled in a separate PR. 👍

frontend/src/components/tags/TagList.vue
sandje00: This comment has been resolved with #50 .

frontend/src/components/profile/ProfileSidebar.vue
sandje00: Current implementation is rendering form fields with `v-for`. It is [not recommendable to use `v-if` with `v-for`](https://learnvue.co/tutorials/vue-for-loop-tips#avoid-use-v-if-in-loops) and the linter does not allow it, either. Changing it any other way it would cause major changes of a component which would belong to another PR.

abasic (approved):

frontend/src/components/profile/ProfileSidebar.vue
Lnola: can it please be added as a task!

frontend/src/components/profile/ProfileSidebar.vue
Lnola: if by any chance the scrollbar right is shown/needs to be displayed `vw` it will mess up scroll left-right. `vh` is ok here, but since we have the overlay being `inset: 0` why not `%` for all

Lnola (approved):
Looks good now!

frontend/src/components/profile/ProfileSidebar.vue
Lnola: I have a solution to this, I'm adding it as a comment where it should be implemented

frontend/src/components/profile/ProfileSidebar.vue
```diff
@@ -0,0 +1,118 @@
...
+  setup() {
+    // TODO: Replace with composables
+    const formInputs = reactive(userProfileFormLabels.formInputs);
+
+    const tags = ref([]);
+
+    onMounted(async () => {
+      const { username, fullName } = useAuthStore().getUser;
+      formInputs.username.value = username;
+      formInputs.fullName.value = fullName;
```
Lnola, line 36: you could do this to remove the `fullName` if it is empty
```
if (formInputs.fullName.value.length === 0) delete formInputs.fullName;
else formInputs.fullName.value = fullName;
```

frontend/src/components/home/HomePage.vue
```diff
@@ -16,6 +18,8 @@ import { roomsSectionsDefault, usersSectionsDefault } from './constants';
 import ChiSidebar from '../common/ChiSidebar.vue';
 import { useAuthStore } from '@/store/authStore';
 import { useRouter } from 'vue-router';
+import ProfileSidebar from '@/components/profile/ProfileSidebar.vue';
```
Lnola, line 21: I'm getting import sorting errors here. No big deal, probably just from merging, but please fix 😄

abasic (approved):

PR #48, opened 2022-08-01
Chat layout

Lnola:
Add the entire chat layout with mock data (connection to the real data will be added in the next PR)

It includes the layout component which uses `ChatNavigation`, `ChatMessages` and `ChatWrite`

`ChatNavigation` is the navigation bar at the top which also includes the tooltip called `ChatDescription` used to add the description data in a more 'hidden' way. It is toggled by pressing the name of the room in the navigation

`ChatMessages` is the chat panel containing all the messages sent to that room. All new messages are added from the bottom to the top. It also includes the `scrollLastIntoView` function which takes a boolean as an argument (default is `false`) to change the behavior to `smooth` if needed (for example: a new message coming in so you want to show scroll and allow the user to stop it)

`ChatWrite` is the textarea that is at the bottom by default. It includes the button to send the message (submit). A new line is added by pressing `return` and submit can also be triggered with either `cmd + return` or `ctrl + return`

Side-effects: 
1. changed `ChiField` error slot under the input to be disabled
2. moved `ChiTextarea` resize function from `onInput` to `onUpdate`


frontend/src/components/chat/ChatWrite.vue
```diff
@@ -0,0 +1,62 @@
...
+  name: 'chat-write',
+  setup() {
+    const input = ref('');
+
+    const handleSubmit = ({ ctrlKey, metaKey }) => {
+      if (ctrlKey || metaKey) {
+        console.log('submit', input.value.trim());
+        input.value = '';
+      } else {
+        console.log('new line');
```
abasic, line 33: I assume these were left here deliberately 😄

frontend/src/components/chat/ChatNavigation.vue
```diff
@@ -0,0 +1,60 @@
+<template>
+  <section class="navigation-container">
+    <div @click="toggleDescriptionActive" class="navigation-heading">
+      <h1 class="navigation-title">Good music</h1>
+      <img class="navigation-arrow" src="@/assets/img/arrow_accent.svg" alt="arrow" />
+    </div>
+
+    <img @click="$emit('toggle-profile')" class="navigation-profile" src="@/assets/img/profile.svg" alt="arrow" />
+
+    <chat-description v-if="isDescriptionActive"></chat-description>
```
abasic, line 10: Generally, `v-if` should be used when something is not often toggled or can not be often toggled. For example, if on page load something needs to be added. `v-show` would be used when the visibility of an element can be changed often or unpredictably (like after a user interaction). The reason for all of this is that `v-if` adds and removes the element from the DOM which can impact performance if done often and on a lot of elements.
In this use case is should not matter much but the recommendation here would be to use `v-show`.

frontend/src/components/chat/ChatMessages.vue
```diff
@@ -0,0 +1,91 @@
...
+import { onMounted, ref } from 'vue';
+
+export default {
+  name: 'chat-messages',
+  setup() {
+    const container = ref(null);
+
+    const scrollLastIntoView = (isSmooth = false) => {
+      const { children } = container.value;
+      children[children.length - 1].scrollIntoView(isSmooth && { behavior: 'smooth' });
```
abasic, line 41: I assume this could have been done by setting the containers `scollTop` value to its `scrollHeight` like this:
```js
element.scrollTop = element.scrollHeight;
```
That way there is no need to fetch the children. But that way the smooth scroll option is not available. But since it is not even used here I think the simpler way might be a better option. It can potentially be made into a directive so it can be used in other places 🤓

abasic (commented):
Just minor comments 👍

sandje00 (commented):
Nothing major, good job!

frontend/src/components/chat/ChatWrite.vue
Lnola: yep! They will be removed soon, but were left as both a reminder and a demo of how it works 😄

frontend/src/components/chat/ChatNavigation.vue
Lnola: good to know, thank you!

frontend/src/components/chat/ChatMessages.vue
Lnola: I did it this way exactly because of the `behavior: 'smooth'` as was written in the PR comments at the top - `(for example: a new message coming in so you want to show scroll and allow the user to stop it)`

I wanted it this way because I think the user browsing old messages would find it frustrating if it kept popping him back down every time a new message arrived. With `behavior: 'smooth'` we can add a button saying something like `New messages ↓` and when clicked it would smoothly transition him to bottom. It can be done without `behavior: 'smooth'` but it wouldn't look as nice 😊 

The idea to make this a directive I love and will look into it in some upcoming PR, probably when we add messages!

Lnola (commented):
Reorder measurements

sandje00 (approved):
👏

abasic (approved):

abasic (approved):

PR #50, opened 2022-08-01
Maintenance/button types

Lnola:
Define default button types. If we change margin, padding or border or some other 'interact with things around it' thing we do it outside the`chi-button` component!


sandje00 (approved):
😎

PR #52, opened 2022-08-02
Tag Management Logic

### LINKED ISSUE #51: Tag management logic

sandje00:
THIS PR:

- Abstract tag handling logic into tag store
- Rewrite current tag handling logic to use tag store (in `TagSelect.vue` and `ProfileSidebar.vue`)

Related issues:
Closes #51


frontend/src/components/profile/ProfileSidebar.vue
```diff
@@ -7,43 +7,41 @@
...
 
-      tags.value = await tagApi.getAll();
+      await setTags();
     });
 
-    const selectTag = tagId => {
-      const tag = tags.value.find(tag => tag.id === tagId);
-      tag.selected = !tag.selected;
+    const closeProfile = async () => {
+      return hasUserChangedTags.value ? saveSelected(emit, 'close-profile') : emit('close-profile');
```
Lnola, line 41: nitpick - does this look cleaner to you?
```
const closeProfile = async () => {
      if (hasUserChangedTags.value) saveSelected();
      emit('close-profile');
};
```
since `emit` gets called in both cases of the conditional statement it makes more sense (to me) to do it this way

Lnola (approved):
Just one minor nitpick you can ignore 😄

frontend/src/store/tagStore.js
```diff
@@ -0,0 +1,44 @@
...
+import { useAuthStore } from '@/store/authStore';
+
+export const useTagStore = defineStore('tag', () => {
+  const userStore = useAuthStore();
+
+  const tags = ref([]);
+
+  const changed = ref(false);
+
+  const getTags = computed(() => {
```
abasic, line 13: Nitpick: since computed properties are used as basic props (even though they are defined as functions) their naming should not follow the naming conventions of functions (unless the computed property returns a function). I assume this was done so that `tags` is not exposed to the outside and can not be changed. If that is the case this could be called `readonlyTags` or something to that effect.

But reading through Pinia's documentation that exposing the state directly is not frowned upon. An issue with Vuex was that it had mutations so the state needed to be changed using it. But it was still possible to change it directly. Since there are no more mutations I do not see exposing state as an issue anymore.

frontend/src/components/profile/ProfileSidebar.vue
abasic: I assume that in this example there should have been an await before the `saveSelected` since it is an `async` function.
I might agree with you but for a different reason. And that is that I am not certain if it needs to be the responsibility of the store action to call the callback. Also, does it always require a callback? If not then more reason to move the emit call to be outside.
See how this will be used and which approach best fits it 😄

abasic (commented):
Minor comments with lots of text 😄

frontend/src/store/tagStore.js
sandje00: For now, since we have used similar logic and naming in `auth` store as well, I would leave `tag` store as it is and make proposed changes in different PR.

Lnola (approved):
Looks good!

PR #54, opened 2022-08-03
Chat layout improvements

### LINKED ISSUE #55: Chat layout improvements

Lnola:
Changed `scrollIntoView` to a custom directive called `vScrollToBottom` which goes on container with the scroll option and by default instantly scrolls to the bottom with the `behavior: auto`. Adding the attribute smooth like so `v-scroll-to-bottom:smooth` adds the `behavior: smooth` option (with a slight delay because of how behavior smooth works 😄)

Updated the way `ChatDescription` can be closed. Now clicking on anything besides the description modal closes it (focus out)

Closes #55


frontend/src/components/chat/ChatNavigation.vue
```diff
@@ -5,9 +5,9 @@
       <img class="navigation-arrow" src="@/assets/img/arrow_accent.svg" alt="arrow" />
     </div>
 
-    <img @click="$emit('toggle-profile')" class="navigation-profile" src="@/assets/img/profile.svg" alt="arrow" />
+    <img @click="$emit('toggle-profile')" class="navigation-profile" src="@/assets/img/profile.svg" alt="profile" />
 
-    <chat-description v-show="isDescriptionActive"></chat-description>
+    <chat-description v-if="isDescriptionActive" @toggle-description="toggleDescriptionActive"></chat-description>
```
Lnola, line 10: this had to be changed because of the `v-autofocus` on `ChatDescription`

frontend/src/directives/vScrollToBottom.js
```diff
@@ -0,0 +1,18 @@
...
+  mounted: (el, binding) => {
+    if (binding.value === undefined || binding.value) {
+      let timeout = 0;
+
+      const isSmooth = binding.arg === 'smooth';
+      if (isSmooth) timeout = 100;
+
+      setTimeout(() => {
+        scrollIntoView(el.children[el.children.length - 1], isSmooth);
+      }, timeout);
```
Lnola, line 13: `setTimeout` is not the ideal solution so if you have a better one I'm all ~~ears~~ eyes 👂👁️ 👁️👂

abasic: Maybe [nextTick](https://vuejs.org/api/general.html#nexttick)?

Lnola: Tried both that and [watchEffect](https://vuejs.org/api/reactivity-core.html#watcheffect), but unfortunately neither of them work in directives 😢

sandje00 (approved):

frontend/src/directives/vScrollToBottom.js
```diff
@@ -0,0 +1,18 @@
...
+const vScrollToBottom = {
+  mounted: (el, binding) => {
+    if (binding.value === undefined || binding.value) {
+      let timeout = 0;
+
+      const isSmooth = binding.arg === 'smooth';
+      if (isSmooth) timeout = 100;
+
+      setTimeout(() => {
+        scrollIntoView(el.children[el.children.length - 1], isSmooth);
```
abasic, line 12: Hint: there is a newer way of finding the last child in an array. It depends on the [at](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/at) method existing on the array. The newest Chrome versions should support it.

frontend/src/directives/vScrollToBottom.js
```diff
@@ -0,0 +1,18 @@
+import scrollIntoView from '@/utils/scrollIntoView';
+
+const vScrollToBottom = {
+  mounted: (el, binding) => {
+    if (binding.value === undefined || binding.value) {
```
abasic, line 5: Nitpick: since the entire method is a single if, maybe an early return could be leveraged to make the code more readable.

frontend/src/components/chat/ChatDescription.vue
```diff
@@ -15,6 +15,12 @@
 <script>
 export default {
   name: 'chat-description',
+  setup(_props, { emit }) {
+    const handleFocusOut = ({ relatedTarget }) => {
+      !relatedTarget && emit('toggle-description');
```
abasic, line 20: Nitpick: I believe a more standardized way of writing this would be:
```js
if (!relatedTarget) emit('toggle-description');
```
Logical AND and OR should, in my eyes, be used when leveraging their [short-circuit evaluation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Logical_AND#short-circuit_evaluation) (did not know it was called that way) to handle more complex logical chaining. Even then I am not certain if they should be used instead of breaking something into multiple lines since it should be easier to read the code if it were multiline. A rule I have been following in recent times is to use them if it really needs to be one line.

frontend/src/components/chat/ChatMessages.vue
```diff
@@ -29,23 +29,11 @@
...
-      const { children } = container.value;
-      children[children.length - 1].scrollIntoView(isSmooth && { behavior: 'smooth' });
-    };
-
-    onMounted(() => {
-      scrollLastIntoView();
-    });
-
-    return { container };
+    const scrollBehavior = '';
```
abasic, line 35: Does this need to be defined or can the directive work without it?

abasic (commented):
Minor comments and nitpicks

frontend/src/directives/vScrollToBottom.js
Lnola: I saw that but forgot about it. Support fits with our project, I'll use it!

Lnola: image

Lnola: returning to the old since even my browser doesn't support it 😅

frontend/src/components/chat/ChatDescription.vue
Lnola: I agree with you 100%, this stayed because I originally put directly inside template without the handler but then later on moved it to `setup()` and it remained so

frontend/src/components/chat/ChatMessages.vue
Lnola: It can work without it but stayed because it will be used later to change the scroll behavior on update!

abasic (approved):

PR #41, opened 2022-07-28
Sidebar - rooms and users

Lnola:
Created the main sidebar component. Currently the displayed data is coming from `constanst.js` but in the future will be coming from the database.

The common `ChiSidebar` component takes two arguments as it's primary data - section content and footer. Content are the displayed `ChiCards` and the footer is the bottom `switcher`. 

A side-effect of this PR is the deletion of the `HomeTemp` component and the start of work on `home` components and screens

Closes #42

PR #58, opened 2022-08-03
Sidebar improvements

### LINKED ISSUE #56: Sidebar improvements

Lnola:
- Main change is moving the default route from `Home` to `Rooms`
- Created new nested routes for the sidebar for displaying rooms and users online
- Separated `ChiSidebar` section mapping into two new components (`ChiSidebar` is just a sidebar with a footer without them)
- Moved the room and user mapping logic from `HomePage` to the new components
- Now includes tags in `getAll` rooms
- Added `roomStore` with the `selectedRoom` inside it
- Add `useTabs` composable

Closes #56


frontend/src/store/roomStore.js
```diff
@@ -0,0 +1,17 @@
+import { computed, ref } from 'vue';
+import { defineStore } from 'pinia';
+import { room as roomApi } from '@/api';
+import { useRoute } from 'vue-router';
+
+export const useRoomStore = defineStore('room', () => {
+  const rooms = ref([]);
+  const route = useRoute();
+
+  const selectedRoom = computed(() => rooms.value.find(({ id }) => id === Number(route.params.id)));
```
abasic, line 10: I have leveraged [this](https://router.vuejs.org/guide/essentials/passing-props.html) functionality in Vue 2 to make change the type of a route param when passing it as a prop to a component. This is not a component but there might be something similar that can be done so that there is no need to manually convert the param into a number here.

frontend/src/components/sidebar/constants.js
```diff
@@ -0,0 +1,94 @@
+const roomsSectionsDefault = [
+  {
+    label: 'Recommended rooms',
+    maxHeight: '30vh',
+    elements: [],
+  },
+  {
+    label: 'All rooms',
+    maxHeight: 'calc(100vh - 30vh - 6 * 24px - 1.5rem - 1.5rem - 2px - 5vh)',
```
abasic, line 9: This is interesting 😄 . Can this be avoided and be in a class somewhere and less complex?

frontend/src/components/sidebar/SidebarUsers.vue
```diff
@@ -0,0 +1,57 @@
...
+import { computed, reactive } from 'vue';
+import ChiCard from '@/components/common/ChiCard.vue';
+import ChiSidebar from '@/components/common/ChiSidebar.vue';
+import { usersSectionsDefault } from './constants';
+
+export default {
+  name: 'sidebar-rooms',
+  setup() {
+    const sections = reactive(usersSectionsDefault);
+    const filteredSections = computed(() => sections.filter(({ elements }) => elements.length > 0));
```
abasic, line 22: This could be a personal preference on your part. If it is, please ignore this comment. But there is no need to test if the length is larger than 0. Zero is falsy so just returning `elements.length` is enough.

frontend/src/store/roomStore.js
```diff
@@ -0,0 +1,17 @@
...
+
+  const setRooms = async () => {
+    if (rooms.value.length === 0) rooms.value = await roomApi.getAll();
```
abasic, line 13: Similar to the other comment related to `length` but just here the `length` can be negated.

backend/src/room/room.controller.ts
```diff
@@ -1,9 +1,9 @@
 import { Request, Response } from 'express';
-import { Room } from 'shared/database';
+import { Room, Tag } from 'shared/database';
 
 const getAll = async (req: Request, res: Response) => {
-  const data = await Room.findAll();
-  return res.json(data);
+  const rooms = await Room.findAll({ include: { model: Tag, through: { attributes: [] } } });
```
abasic, line 5: Since this might be a common query, it could be useful to move it to a scope.

frontend/src/components/sidebar/SidebarRooms.vue
```diff
@@ -0,0 +1,85 @@
...
+    const filteredSections = computed(() => sections.filter(({ elements }) => elements.length > 0));
+
+    const handleSelect = id => {
+      router.push({ name: 'Rooms', params: { id } });
+    };
+
+    onMounted(async () => {
+      await store.setRooms();
+
+      sections[1].elements = store.rooms.map(({ id, title }) => ({
```
abasic, line 43: Is this placeholder code that will be replaced with something from the backend?

abasic (commented):
Just minor comments and questions.

frontend/src/router/index.js
```diff
@@ -30,13 +44,25 @@ const router = createRouter({
...
-    return { name: 'Home' };
+    return defaultRoute;
+  }
+
+  if (isHomeRoute) {
+    return defaultRoute;
+  }
+
+  if (isRoomsRoute && !to.params.id) {
+    return defaultRoute;
```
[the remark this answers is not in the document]
Lnola: I used it where I could and thought it made sense for now

frontend/src/store/roomStore.js
Lnola: I haven't found any way of doing it except writing a helper function that would return the param as a `Number`. That would be a bit overkill in my opinion

frontend/src/components/sidebar/constants.js
Lnola: This is 💩 but unfortunately it cannot be changed easily. I played it with for a few days actually (this was already in main since #41, I just replaced the file location so it came up as a change now). I couldn't find a better way of doing it with css. It can be moved to a class now (previously I couldn't even do that 😄) so I'll do that

frontend/src/components/sidebar/SidebarUsers.vue
Lnola: yep, personal preference thing, I like having it declared so it is immediately clear what is being checked. Since I've seen @sandje00 using the same thing you suggested I could switch this no problem if you think it makes sense to do so

frontend/src/components/sidebar/SidebarRooms.vue
Lnola: no, this maps the room data into something the default `ChiCard` can use, something that, in my opinion, has nothing to do with the backend. The backend returns the data as it is, defined in the model, and on the frontend I save it in the `roomsStore`. After that I use that original data in the whole app along with mapping it into something `ChiCard` can read

frontend/src/components/sidebar/SidebarUsers.vue
sandje00: Personally I tend to leverage falsy nature of 0, but I don't see this being a major problem if it stays so, therefore, you don't have to change it.

frontend/src/components/sidebar/constants.js
abasic: We can leave it for now and I will look into it and see what I come up with 🤓

frontend/src/components/sidebar/SidebarRooms.vue
abasic: While reviewing, I was wondering if this was the right place for it. If it should be here or in the store. So I wanted to know if it was something that will change soonish or not. I am still not certain if it should be in the store or not so lets leave it here for now😄

frontend/src/components/sidebar/SidebarUsers.vue
abasic: If it is personal preference then so be it. My comment then would be for you two to consolidate your approaches so that it is more consistent.

frontend/src/router/index.js
```diff
@@ -9,6 +11,19 @@ const routes = [
     path: '/',
     name: 'Home',
     component: Home,
+    redirect: { name: 'Rooms' },
+    children: [
```
abasic, line 15: tl;dr this comment can be ignore 😄 

An approach sometimes that is taken is to have a child that has the same path as the parent. Something like this:
```js
{
  path: '/',
  component: Home,
  children: [
    {
      path: '',
      name: 'Home',
      redirect: { name: 'Rooms' }
    },
    ...
  ]
```
Usually, the parent does not have a name since a child always needs to be loaded. Redirecting a user to just a parent would cause an error. So the approach is for the first child to have the same path as the parent. Here, since it is just a redirect and there is no child component, it can be like this.
But now, looking at it, another approach could be for the child in the code above to use the `SidebarRooms` component. That way, there would be no need for the second child to have an optional parameter. Something like this:
```js
{
  path: '/',
  component: Home,
  children: [
    {
      path: '',
      name: 'Home',
      component: SidebarRooms
    },
    {
      path: 'rooms/:id',
      name: 'Rooms',
      component: SidebarRooms
    }
  ]
```
I am personally not a fan of optional parameters since I find them to be more prone to errors. And I have a preference for being as explicit as possible in code.

frontend/src/router/index.js
```diff
@@ -9,6 +11,19 @@ const routes = [
...
+      {
+        path: 'rooms/:id?',
```
abasic, line 17: Something that just crossed my mind is that there might be a need for [passing props to route components](https://router.vuejs.org/guide/essentials/passing-props.html). So keep that in mind in the future since it might be useful.

abasic (approved):
Just two minor, more educational comments.

frontend/src/router/index.js
Lnola: the optional parameter is there just for the redirect since, without it, the redirect wouldn't work. Trying out the different options, I couldn't fit them nicely so I think it's better to think of this as part of some other PR which will focus on cleaning up routes and defining default redirects/not found page since we don't have that now (this here is just proofing so it works and never breaks the dev server)

sandje00 (approved):
🚀

PR #67, opened 2022-08-10
Basic socket connection

Lnola:
Created the basic socket connection setup with socket.io. First major change - `const server = http.createServer(app);` - app was replaced with `httpServer` to allow a realtime connection. All socket communication is done through what I called `actions`. The connection has an authentication middleware based on sending the auth token with the `connection` action and the auth middleware intercepting it and extending the `socket` object with the parameter `username`.  

The actions are (on the FE listeners become emits and emits become listeners):

- connect and disconnect
- join and leave room
- emitting on number of users in each room

`Connect` creates the connection and defines the socket with its id. That object is used everywhere for communicating with the connected user. Disconnect is triggered when the connection is closed. `Join` room is simple - it defines that the user socket is now part of the socket defined object room (not related to our own `room` object). That room object is used to emit events to all users in some room with `io.to('room').emit(...)`. Leave is self explanatory, it defines the socket leaving the room but continuing with the connection. `Emitting active` users comes in two parts - emitting the number of active users for all rooms (which is triggered on `socket connection` and `socket disconnect`) while emitting the active users for the room user joined/left emits only the room user left and the one they joined (in that order, triggered on `leave` and `join`)

Packages added: 

- socket.io (BE)
- socket.io-client (FE)
- universal-cookie (FE)
- @vueuse/integrations (FE)

Config is simple - socket targets `http://localhost:3000/socket.io...` while the proxy redirects all `/socket` calls to `http://backend:3001`

Closes: #49


frontend/src/router/index.js
```diff
@@ -12,11 +12,13 @@ const routes = [
     name: 'Home',
     component: Home,
     redirect: { name: 'Rooms' },
+
```
abasic, line 15: Lint: why the empty line between two object props?

frontend/src/router/index.js
```diff
@@ -12,11 +12,13 @@ const routes = [
...
     children: [
       {
         path: 'rooms/:id?',
         name: 'Rooms',
         component: SidebarRooms,
+        props: route => ({ routeId: Number(route.params.id) }),
```
abasic, line 21: Nitpick: since `id` does not need to exist, `Number(undefined)` returns `NaN`. I believe it would be a more standard approach if the prop was set to `null` if the route param did not exist.

backend/src/app.ts
```diff
@@ -22,8 +25,11 @@ app.use('/api', router);
   }
 })();
 
-app.listen(port, function () {
+const server = http.createServer(app);
```
abasic, line 28: I have nothing against adding the `http` package and using it to create the server. But the PR comment mentioned it needed to be added to allow real-time communication. Looking at [this](https://tsh.io/blog/socket-io-tutorial-real-time-communication/) specific example, it might not be needed. Socket IO's documentation [mentions](https://socket.io/docs/v4/server-initialization/#with-express) using the `http` server with Express but I have not been able to find explicitly why that is. Do you know why that is or is the example I linked incorrect?

backend/src/shared/helpers/sockets.ts
```diff
@@ -0,0 +1,27 @@
+import { io } from 'sockets';
+import { ISocket } from 'shared/sockets/types';
+
+function getNumberOfUsersInAllRooms() {
+  const arr = Array.from(io.sockets.adapter.rooms);
+  const filtered = arr.filter(room => !room[1].has(room[0]));
```
abasic, line 6: Can this be explained a little bit? It is apparent logically what it is trying to do but what the values hold is not. Can something like this be done:
```js
const filtered = arr.filter(([var1, var2]) => !var2.has(var1));
```
Where `var1` and `var2` are more descriptive names that describe what values they hold.

backend/src/shared/helpers/sockets.ts
```diff
@@ -0,0 +1,27 @@
...
+
+  return filtered.map(i => ({ roomId: i[0], numberOfUsers: i[1].size }));
```
abasic, line 8: Same as above.

backend/src/shared/helpers/sockets.ts
```diff
@@ -0,0 +1,27 @@
...
+
+const getSocketsInRoom = (roomId: string) => io.sockets.adapter.rooms.get(roomId);
+
+const getNumberOfUsersInRoom = (roomId: string) => getSocketsInRoom(roomId)?.size || 0;
+
+const getActiveUsersInRoom = (roomId: string) => {
+  const sockets = getSocketsInRoom(roomId);
+  const activeUsernames: Array<string | undefined> = [];
+
+  sockets?.forEach(socketId => {
```
abasic, line 19: Should `sockets` not always be an array? If `getSocketsInRoom` always returns an array then this could have been a `map` instead since that fits more to what is being done.

backend/src/sockets/actions.ts
```diff
@@ -0,0 +1,38 @@
...
+    console.log(`${socket.username} left room ${roomId}`);
+  };
+
+  emitNumberOfUsersInAllRooms = (socket?: ISocket) => {
+    if (!socket) {
+      io.emit('total-number-of-users', getNumberOfUsersInAllRooms());
+      return;
+    }
+
+    socket.emit('total-number-of-users', getNumberOfUsersInAllRooms());
```
abasic, line 30:
```suggestion
const emitter = socket || io;
emitter.emit('total-number-of-users', getNumberOfUsersInAllRooms());
```
I do not know if this will cause any `TypeScript` issues.

frontend/src/utils/mapNumberOfUsersToRooms.js
```diff
@@ -0,0 +1,15 @@
+const mapNumberOfUsersToRoom = (roomId, numberOfUsers, roomStore) => {
+  roomStore.rooms.find(({ id }) => id === roomId).activeUsers = numberOfUsers;
```
abasic, line 2: Nitpick: personally I am not a fan of a single line doing multiple things. In my mind, it can lead to code that is difficult to read. This line is not that difficult since it is straight forward but I would still recommend storing the room in a variable.
```js
const room = roomStore.rooms.find(({ id }) => id === roomId);
room.activeUsers = numberOfUsers;
```

frontend/src/utils/mapNumberOfUsersToRooms.js
```diff
@@ -0,0 +1,15 @@
...
+};
+
+const mapNumberOfUsersToAllRooms = (totalNumberOfUsers, roomStore) => {
+  roomStore.rooms.forEach(({ id }) => {
+    let instance = totalNumberOfUsers.find(el => el.roomId === id);
```
abasic, line 7: Nitpick: `el` might not be the best name for the variable in the method. Maybe `data` (very generic I know) or something like that. Personally, I associate `el` with element (either HTML or something like that on the frontend).

frontend/src/socket/chat.js
```diff
@@ -0,0 +1,39 @@
...
+  listenToNumberOfUsers(roomStore) {
+    if (this.socket) {
+      this.socket.on('number-of-users', ({ roomId, numberOfUsers }) =>
+        mapNumberOfUsersToRoom(roomId, numberOfUsers, roomStore),
+      );
+    }
+  }
+}
+
+export default new ChatService();
```
abasic, line 39: Something very cute: since every method checks if `this.socket` exists, this could be made into a proxy that checks that before the calls to most of the methods. Maybe even calling `createConnection` if it does not exist.

frontend/src/components/sidebar/SidebarRooms.vue
```diff
@@ -17,39 +17,58 @@
...
 import ChiSidebar from '@/components/common/ChiSidebar.vue';
 import { roomsSectionsDefault } from './constants';
 import { useRoomStore } from '@/store/roomStore';
 import { useRouter } from 'vue-router';
+import { useSocketStore } from '@/store/socketStore';
 
 export default {
   name: 'sidebar-rooms',
-  setup() {
+  props: { routeId: { type: Number, default: 1 } },
```
abasic, line 30: Nitpick: maybe spread this so it is slightly easier to read.
```js
props: {
  routeId: { type: Number, default: 1 }
}
```

frontend/src/components/sidebar/SidebarRooms.vue
```diff
@@ -8,7 +8,7 @@
           :key="element.id"
           @click="handleSelect(element.id)"
           v-bind="element"
-          :selected="store.selectedRoom.id === element.id"
+          :selected="roomStore.selectedRoom.id === element.id"
```
abasic, line 11: Nitpick: maybe just expose `selectedRoom` to the template since it is the only thing from the `roomStore`.

frontend/src/components/sidebar/SidebarRooms.vue
```diff
@@ -17,39 +17,58 @@
...
 
-      sections[1].elements = store.rooms.map(({ id, title }) => ({
-        id,
-        iconLeft: 'globe.svg',
-        textLeft: title,
-        textRight: String(0),
-        iconRight: 'profile_darker.svg',
-      }));
+    watchEffect(() => {
+      mapSections();
```
abasic, line 68: How often does this `watchEffect` get triggered in a lifecycle of the component? I am generally weary of using watchers since it can be difficult to figure out what triggered them. So I like to figure out if there is a single event that always triggers them or a more linear (might not be the right word), predictable way for the data to flow.

abasic (commented):
I believe mostly nitpicks and questions 👍

frontend/src/router/index.js
Lnola: image

Lnola: image

frontend/src/router/index.js
Lnola: `id` is optional only because that way I could register it to redirect to `rooms/1`. `/room` without the `id` will never be accessed so if it is `NaN` for a second before the redirect it should not be a problem

backend/src/app.ts
Lnola: I initially tried doing it without the `http` created server and got this error while creating the new socket connection (`io`)

```
No overload matches this call.
  Overload 1 of 3, '(opts?: Partial<ServerOptions> | undefined): Server<DefaultEventsMap, DefaultEventsMap, DefaultEventsMap, any>', gave the following error.
    Argument of type 'Application' is not assignable to parameter of type 'Partial<ServerOptions>'.
      Types of property 'path' are incompatible.
        Type '() => string' is not assignable to type 'string'.
  Overload 2 of 3, '(srv?: number | import("http").Server | import("https").Server | undefined, opts?: Partial<ServerOptions> | undefined): Server<...>', gave the following error.
    Argument of type 'Application' is not assignable to parameter of type 'number | Server | Server | undefined'.
      Type 'Application<Record<string, any>>' is missing the following properties from type 'Server': closeAllConnections, closeIdleConnections, addContext, getTicketKeys, and 17 more.
  Overload 3 of 3, '(srv: number | import("http").Server | import("https").Server | Partial<ServerOptions> | undefined, opts?: Partial<ServerOptions> | undefined): Server<...>', gave the following error.
    Argument of type 'Application' is not assignable to parameter of type 'number | Server | Server | Partial<ServerOptions> | undefined'.
      Type 'Application<Record<string, any>>' is not assignable to type 'Server'.ts(2769)
```

The example you sent works (I tried it now) and the reason it works is because `app.listen(...)` returns an object of type `http.Server`. That means that instead of doing what I did, we could do something like this and it would have the same result

```
const server = app.listen(port, function () {
  console.log(`App is listening on port ${port} !`);
});

socketConnection(server);
```

Since I don't see the difference and I've done it following the documentation, I don't see the benefit of switching to this, but its good to know this can be done!

backend/src/shared/helpers/sockets.ts
Lnola: forgot about array destructuring! I'll add more descriptive names

backend/src/shared/helpers/sockets.ts
Lnola: since the room may not exist (according to TS even though that is not possible with the current state of things...) it may return `undefined` and because of that it had to be `?`

However, I can do this 
```
const getSocketsInRoom = (roomId: string) => io.sockets.adapter.rooms.get(roomId) || new Set<string>();
```
which allows me to do what you suggested!

Lnola: I changed the `getSocketsInRoom` to return an `Array` instead of a `Set`. That way I can map it and it works fine!

backend/src/sockets/actions.ts
Lnola: I was sure something like that didn't even work in `TS` but apparently it does! Love it! ❤️

frontend/src/components/sidebar/SidebarRooms.vue
Lnola: 🧈

frontend/src/components/sidebar/SidebarRooms.vue
Lnola: It cannot be easily done without losing reactivity or adding a bunch of code to make it reactive. If you have suggestions to achieve the same thing without complicating, I'm of the opinion that this would be better this way

frontend/src/components/sidebar/SidebarRooms.vue
Lnola: I am aware of the possible bad side-effects of using watchers but the problem here is that this watcher listens for emits from the socket so it cannot be triggered by anything else (that I'm aware of) but the watchEffect.

Just for quick context - this maps the number of users in rooms which is data that gets emitted when a new user on another connection creates the connection

frontend/src/router/index.js
abasic: Is a room with `id` that equals 1 a default room with all tags or something equivalent to a general chat room? If not, why would the user be redirected specifically to it other than the room having the first `id` generated?

frontend/src/components/sidebar/SidebarRooms.vue
abasic: I'll try something later but just to confirm something like this does not work:
```js
return { selectedRoom: roomStore.selectedRoom, sections, filteredSections, handleSelect };
```

frontend/src/router/index.js
Lnola: yes, the idea was room `id = 1` would be like `general` on slack - and in a real release (with room CRUD) would be seeded with `id` of 1

frontend/src/components/sidebar/SidebarRooms.vue
Lnola: nope, tried it and it doesn't work (it causes errors which can be fixed by adding a `?` but it is still not reactive)

frontend/src/components/sidebar/SidebarRooms.vue
abasic: I am not saying to do this since it is more of the top of my head and it might be more problematic but, a combination of multiple computed might be useful to solve this. An approach would something like this:
```js
const sections = computed(() => {
  const recommended =  {
    ...roomsSectionsDefault[0],
    elements: []
  };
  const all = {
    ...roomsSectionsDefault[1],
    elements: roomStore.rooms.map(({ id, title, activeUsers = 0 }) => {
      return {
        id,
        iconLeft: 'globe.svg',
        textLeft: title,
        textRight: String(activeUsers),
        iconRight: 'profile_darker.svg',
      };
     })
  };

  return [recommended, all];
});
const filteredSections = computed(() => sections.filter(({ elements }) => elements.length));
```

Something that could be done is have two computeds for both sections and then merge them in the `filteredSections` computed. Something like this:
```js
const recommendedRooms = computed(() => {
  return  {
    ...roomsSectionsDefault[0],
    elements: []
  };
});

const alllRooms = computed(() => {
  ...roomsSectionsDefault[1],
  elements: roomStore.rooms.map(({ id, title, activeUsers = 0 }) => {
    return {
      id,
      iconLeft: 'globe.svg',
      textLeft: title,
      textRight: String(activeUsers),
      iconRight: 'profile_darker.svg',
    };
   })
});

const filteredSections = computed(() => [recommendedRooms, allRooms].filter(({ elements }) => elements.length));
```
This way not all of them are recalculated when a change happens. Look into it and see if you like the approach.

sandje00 (approved):
Everything looks good!

frontend/src/components/sidebar/SidebarRooms.vue
Lnola: Love the second approach, went with it, went a little further and set up the mapping function for @sandje00 when we connect the recommendedRooms

frontend/src/components/sidebar/SidebarRooms.vue
```diff
@@ -18,38 +18,61 @@
...
+  props: {
+    routeId: { type: Number, default: 1 },
+  },
+  setup(props) {
     const router = useRouter();
-    const store = useRoomStore();
+    const roomStore = useRoomStore();
+    const socketStore = useSocketStore();
 
     const sections = reactive(roomsSectionsDefault);
```
abasic, line 39: I do not think these need to be reactive anymore since they will not be directly changed.

frontend/src/components/sidebar/SidebarRooms.vue
```diff
@@ -18,38 +18,61 @@
...
-    const filteredSections = computed(() => sections.filter(({ elements }) => elements.length));
+
+    const recommendedRooms = computed(() => {
+      return {
+        ...sections[0],
```
abasic, line 43: Nitpick: since the sections become an array in the `fillteredSections` computed, they can now be an object with prop names like `recommended` and `all`. That way they are not accessed here through the index but with a descriptive prop name. (this is something I wanted to comment since I first saw it but I was waiting for the right moment 🤓 )

frontend/src/socket/index.js
```diff
@@ -0,0 +1,12 @@
+import ChatService from './chat';
+
+const chatServiceTarget = new ChatService();
+
+const chatServiceHandler = {
+  get(target, prop) {
+    if (!target.socket && prop !== 'createConnection') return target.createConnection();
+    if (typeof target[prop] === 'function') return target[prop].bind(target);
```
abasic, line 8: Nitpick: maybe have a default return here that returns the data requested from the original object.

abasic (approved):
Very minor nitpicks so I am approving the PR 👍

frontend/src/components/sidebar/SidebarRooms.vue
Lnola: this was an array because originally (when it was all in one component - both users and rooms) it was flexible as to how many of them (the sections) there were. Since these have been split up into multiple specialised components that flexibility isn't a factor anymore so I agree with you, these indexes will be changed with proper names. 

The time of :poop: code `constants` and `sidebar` is finally over!!

frontend/src/socket/index.js
Lnola: I still do not understand this properly so for now I'll return the `target` object since it makes the most sense for now and it works. I'm adding this as an `improvement` and will revisit it later.

abasic (approved):

sandje00 (approved):

PR #61, opened 2022-08-04
Room improvements

Lnola:
This PR brings the display of the selected room data inside `ChatLayout` - title in `ChatNavigation` and description  + tags in `ChatDescription`

Closes #59

PR #69, opened 2022-08-12
Messaging

Lnola:
Add messaging to sockets and to the database. Major additions and changes:
Added message sending to rooms. User `emits` his message, the server listens and `re-emits` the message to all the users in the same room but also stores the `message` in the database. 
On the frontend side, the user joins a room and immediately fetches **all messages** (infinite scroll TBA) from that room and gets scrolled to the bottom. When a message arrives, the users screen

1. jumps to the bottom immediately if he is at the last message
2. scrolls to the bottom if he is scrolled up reading older messages

Along with that, the more minor changes:
- Updated the messages `getAll` to `getAllByRoom` and protected the route. 
- Along with that the `messages model` has been modified with `text` instead of `content` attribute (changed by the latest migration)
- `ISocket` has been extended with `userId`
- Added `dayjs` as a dependency and created the util function to format the date for display on the frontend messages
- Changed `v-scroll-to-bottom` `mounted` to `updated` and fixed bug causing it to crash if the array of elements was empty

Closes: #68


frontend/src/components/chat/ChatMessages.vue
```diff
@@ -1,40 +1,50 @@
...
+        <p class="message-text">{{ message.text }}</p>
+        <span class="message-timestamp">{{ message.createdAt }}</span>
       </div>
     </article>
   </section>
 </template>
 
 <script>
+import { reactive, ref, watchEffect } from 'vue';
+import { useRoomStore } from '@/store/roomStore';
```
abasic, line 15: The code is throwing an error here because the `roomStore` module only has a default export.

backend/src/sockets/actions.ts
```diff
@@ -21,6 +22,15 @@ class SocketActions {
     console.log(`${socket.username} left room ${roomId}`);
   };
 
+  emitMessage = (socket: ISocket, roomId: string, message: string) => {
+    const { userId, username } = socket;
+    const text = message;
+    const createdAt = new Date();
+
+    io.to(roomId).emit('message', { username, text, createdAt });
+    Message.create({ userId, roomId, text, createdAt });
```
abasic, line 31: This code does not wait for the message to be stored. What if the message does not get created for some reason? 👀

frontend/src/socket/chat.js
```diff
@@ -15,6 +16,18 @@ class ChatService {
     this.socket.emit('exit', roomId);
   }
 
+  emitMessage(roomId, message) {
+    if (this.socket) {
```
abasic, line 20: I do not think this `if` and the one in the method below this one are necessary anymore.

frontend/src/components/chat/ChatMessages.vue
Lnola: I merged with main without checking this, it is happening because #61 changed the default store export. I'll fix all issues as soon as possible!

frontend/src/components/chat/ChatMessages.vue
```diff
@@ -1,40 +1,50 @@
...
-    const scrollBehavior = '';
-    return { scrollBehavior };
+    const scrollBehavior = reactive({ smooth: false, timeout: 0 });
+
+    const roomStore = useRoomStore();
+    const socketStore = useSocketStore();
+    const messages = ref([]);
+
+    watchEffect(() => {
+      if (!socketStore.socket) return;
```
abasic, line 28: Is this `if` still necessary? If it is, maybe a store action could have been added as a wrapper around `listenToMessage` that does the check. That way the component does not need to worry about the constraints of listening to messages.

frontend/src/components/chat/ChatMessages.vue
```diff
@@ -1,40 +1,50 @@
...
+      messages.value = roomStore.messages;
+    });
+
+    let scroll;
+    const handleScroll = ({ target: { offsetHeight, scrollHeight, scrollTop } }) => {
+      clearTimeout(scroll);
+
+      scroll = setTimeout(function () {
+        if (offsetHeight + scrollTop + 0.5 < scrollHeight) scrollBehavior.smooth = true;
+        else scrollBehavior.smooth = false;
```
abasic, line 43:
```js
const isUnderThreshold = offsetHeight + scrollTop + 0.5 < scrollHeight;
scrollBehavior.smooth = isUnderThreshold;
```

Nitpick/lint: Maybe a better name could be used for the variable but I believe this way code is easier to understand. The variable could be skipped but I wanted to see even for myself if I like it more with it or without it 😄

frontend/src/store/roomStore.js
```diff
@@ -13,11 +15,20 @@ const useRoomStore = defineStore('room', () => {
     if (!rooms.value.length) rooms.value = await roomApi.getAll();
   };
 
+  const setMessages = async () => {
+    if (selectedRoom.value) {
+      messages.value = (await messageApi.getAllByRoom(selectedRoom.value.id)).map(message => ({
+        ...message,
+        createdAt: formatDateForDisplay(message.createdAt),
+      }));
+    }
```
abasic, line 24:
```js
    const roomId = selectedRoom.value?.id;
    if (!roomId) return;

    const roomMessages = await messageApi.getAllByRoom(roomId);
    messages.value = roomMessages.map(message => ({
        ...message,
        createdAt: formatDateForDisplay(message.createdAt),
      }));
    }
```

Since only the `id` is needed, this seemed like an easier-to-read approach.

abasic (changes requested):

backend/src/sockets/actions.ts
Lnola: Nice catch and a big oversight on my part. I added handling now!

frontend/src/components/chat/ChatMessages.vue
Lnola: Makes sense, adding the wrapper inside socketStore. I'll also add it as a task to check if all socket actions should be done that way (through the store instead of accessing the socket directly)

frontend/src/components/chat/ChatMessages.vue
Lnola: Love it!

frontend/src/store/roomStore.js
Lnola: Much cleaner, nice!

abasic (approved):

sandje00 (approved):

PR #74, opened 2022-08-16
Active users sidebar

Lnola:
Now displays the currently active users in each room

Additions:
- emitters and listeners for getting users online
- mapping users online
- changing the `constants.js` file and updating the code using it

The major side-effects are:
- add `id` to the `UsersOnline` route
- add `activeUsers` to the store
- reroute `UsersOnline` if `id` is undefined
- switch tabs now switches to the correct `id`
- users online sections mapping changed to follow new structure but still allow future section additions 
- move socket initiation from `SidebarRooms` to `HomePage`
- add `routeId` as prop to `HomePage`

Closes: #72


frontend/src/socket/chat.js
```diff
@@ -37,6 +37,14 @@ class ChatService {
       mapNumberOfUsersToRoom(roomId, numberOfUsers, roomStore.recommendedRooms);
     });
   }
+
+  listenToActiveUsersInRoom(roomStore) {
+    if (this.socket) {
```
abasic, line 42: Is this `if` needed?

frontend/src/router/index.js
```diff
@@ -50,7 +51,7 @@ const router = createRouter({
 router.beforeEach((to, from) => {
   const isLoggedIn = useAuthStore().isLoggedIn;
   const isAuthRoute = to.name === 'Auth';
-  const isRoomsRoute = to.name === 'Rooms';
+  const isHomeRoute = to.name === 'Rooms' || to.name === 'UsersOnline';
```
abasic, line 54: Hint: a different way to do this would be to leverage [meta fields](https://router.vuejs.org/guide/advanced/meta.html). By setting on these two routes a meta field `isHomeRoute` then it could just be checked in the `if` on line 66. And with this approach it is easier to maintain because then it is just a matter of changing the configuration (route definition) instead of changing the logic code (`beforeEach`).

frontend/src/socket/chat.js
Lnola: Nope, it's a remnant I forgot to change during rebasing. Removing it!

frontend/src/components/sidebar/SidebarUsers.vue
```diff
@@ -1,27 +1,44 @@
...
     </section>
   </chi-sidebar>
 </template>
 
 <script>
-import { computed, reactive } from 'vue';
 import ChiCard from '@/components/common/ChiCard.vue';
 import ChiSidebar from '@/components/common/ChiSidebar.vue';
+import { computed } from 'vue';
+import { useRoomStore } from '@/store/roomStore';
```
sandje00, line 16: Maybe this should be imported from `@/store`?

Lnola: Nice catch!

sandje00 (approved):

PR #78, opened 2022-08-16
Add default catchAll redirect

Lnola:
Closes: #77


frontend/src/router/index.js
```diff
@@ -36,6 +36,10 @@ const routes = [
     name: 'Tags',
     component: TagSelect,
   },
+  {
+    path: '/:catchAll(.*)',
```
abasic, line 40: What is the specific reason this is default catch is written this way? From what I know (at least from the older version of `vue-router`), the path could just be `*` and it would work.

Lnola: It is explained [here](https://programmerah.com/solved-vue3-configuration-routing-error-catch-all-routes-must-now-be-defined-using-a-param-with-a-custom-regexp-32886/). I tried it first with `*` but got had the same error as stated in the article, `*` doesn't work anymore

abasic: Well that is stupid 😛

abasic (approved):

sandje00 (approved):

PR #86, opened 2022-08-18
Time tracking

Lnola:
Add tracking user room interaction (time user spent in each room). How it works:

1. Once the user enters a room a new `userRoomInteraction` is created with these params: `{ userId, roomId, joinedAt: new Date(), leftAt: null }`
2. When the user leaves the room or disconnects from the socket (browser refresh/close) `leftAt` is saved as `new Date()`

Along with that `userRoomInteraction` seed has been added for current users

Closes: #85


backend/src/sockets/actions.ts
```diff
@@ -8,18 +9,42 @@ class SocketActions {
...
   };
 
-  onJoinRoom = (socket: ISocket, roomId: string) => {
+  onJoinRoom = async (socket: ISocket, roomId: string) => {
     socket.join(roomId);
     console.log(`${socket.username} joined room ${roomId} with ${getNumberOfUsersInRoom(roomId)} active users`);
+
+    try {
+      const userId = socket.userId as number;
+      createUserRoomInteraction(userId, roomId);
```
abasic, line 30: Is this supposed to be `await`-ed?

Lnola: yep...

abasic (approved):

sandje00 (approved):

PR #87, opened 2022-08-18
Tag Intensity Based Recommendation

### LINKED ISSUE #84: Tag intensity based room recommendation

sandje00:
THIS PR:

- Add `intensity` attribute to `userTags` relation
- Seed intensity values
- Add migration function to drop existing recommendation materialized view and its related database objects and create new one which uses tag intensity for room recommendation, as well as the function to revert that action
- Update `RecommendedRoom` model to match new meterialized view fields

[Tag Intensity Based Room Recommendation Query Explanation](https://github.com/ExtensionEngine/chichet/wiki/Tag-Intensity-Based-Room-Recommendation)

Related issues:
Closes #84


backend/src/room/recommendedRoom.model.ts
```diff
@@ -20,8 +20,9 @@ class RecommendedRoom extends Model {
       title,
       description,
       roomScore: { type: INTEGER },
+      tagsMatched: { type: INTEGER },
       tagsTotal: { type: INTEGER },
-      roomScoreDiff: { type: INTEGER },
+      tagsNumDiff: { type: INTEGER },
```
Lnola, line 25: nitpick: I am not a fan of shortened names (unless there is a reason) since I think it makes the code less readable

backend/src/userTag/userTag.model.ts
```diff
@@ -21,6 +21,11 @@ class UserTag extends Model {
         allowNull: false,
         unique: 'unique_user_id_tag_id',
       },
+
+      intensity: {
+        type: INTEGER,
+        allowNull: true,
```
Lnola, line 27: since intensity will be calculated as a percentage perhaps `decimal` would fit better. That way we don't have to use `round` and we don't have to multiply it by 10

Lnola: we could add a constraint min/max `0 < intensity < 1`  as well

backend/src/shared/database/seeders/20220803123242-user-tags.js
```diff
@@ -1,27 +1,27 @@
...
+      { userId: 4, tagId: 20, intensity: 8 },
+      { userId: 5, tagId: 13, intensity: 4 },
+      { userId: 5, tagId: 14, intensity: 4 },
+      { userId: 5, tagId: 27, intensity: 9 },
+      { userId: 5, tagId: 28, intensity: 8 },
+      { userId: 6, tagId: 9, intensity: 8 },
+      { userId: 6, tagId: 16, intensity: 5 },
+      { userId: 6, tagId: 21, intensity: 9 },
+      { userId: 6, tagId: 26, intensity: 7 },
+      { userId: 6, tagId: 31, intensity: 4 },
```
Lnola, line 24: these need to be updated if intensity `INTEGER`/`DECIMAL` is changed

backend/src/shared/database/migrations/20220817094029-add-intensity-to-user-tag.js
```diff
@@ -0,0 +1,12 @@
+module.exports = {
+  up(queryInterface, Sequelize) {
+    return queryInterface.addColumn('userTags', 'intensity', {
+      allowNull: true,
+      type: Sequelize.INTEGER,
```
Lnola, line 5: this need to be updated if intensity `INTEGER`/`DECIMAL` is changed

abasic (dismissed):

backend/src/shared/database/migrations/20220817203222-add-intensity-to-recommended-rooms.js
```diff
@@ -0,0 +1,179 @@
...
+module.exports = {
+  async up(queryInterface) {
+    await dropTagBasedRecommendation(queryInterface);
+    return createTagIntensityBasedRecommendation(queryInterface);
+  },
+
+  async down(queryInterface) {
+    await dropTagIntensityBasedRecommendation(queryInterface);
+    return createTagBasedRecommendation(queryInterface);
+  },
```
abasic, line 95: Thinking about it, maybe transactions should be used here as well.

abasic (changes requested):

backend/src/shared/database/migrations/20220817094029-add-intensity-to-user-tag.js
```diff
@@ -0,0 +1,21 @@
...
+  },
+
+  async down(queryInterface) {
+    await queryInterface.sequelize.query(`
+      ALTER TABLE "userTags"
+      DROP CONSTRAINT "intensityRange";
+    `);
+    return queryInterface.removeColumn('userTags', 'intensity');
+  },
+};
```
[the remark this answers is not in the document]
sandje00: Since adding transactions to migrations with multiple queries would introduce changes to files that are not part of this PR as well, I will add it as a part of a different PR.

abasic (approved):
I'll approve this but I generally recommend the usage of [managed transactions](https://sequelize.org/docs/v6/other-topics/transactions/#managed-transactions) because `commit` and `rollback` do not need to be called directly but are done automatically.

Lnola (approved):

PR #89, opened 2022-08-19
Messages infinite scroll

Lnola:
This PR adds infinite scroll to messages.

- Rename `getAllByRoom` to `getByRoom` and update it with `limit` and `offset` to allow `load more messages` required for infinite scroll. 
- Update watchers and handling of adding new messages
- Update `handleScroll` with `offsetIndex` increment and block it when it should be stopped
- Implement infinite scroll
- Add custom spinner/loader component

Closes: #88


backend/src/message/message.controller.ts
```diff
@@ -1,11 +1,16 @@
...
-const getAllByRoom = async (req: Request, res: Response) => {
+const getByRoom = async (req: Request, res: Response) => {
+  const limit = Number(req.query.limit);
+  const offset = Number(req.query.offset);
   const { roomId } = req.params;
-  const messages = await Message.findAll({ where: { roomId } });
+
+  const messages = await (
+    await Message.findAll({ where: { roomId }, order: [['id', 'DESC']], limit, offset })
+  ).reverse();
```
abasic, line 11: Since `reverse` is in place method (does not create a new array but mutates the current one) this could have been written like this:
```js
const messages = await Message.findAll({ where: { roomId }, order: [['id', 'DESC']], limit, offset });
messages.reverse();
```

or
```js
const messages = await Message.findAll({ where: { roomId }, order: [['id', 'DESC']], limit, offset });
return res.json(messages.reverse());
```
Also, I do not believe the await before the brackets was necessary. It implies the `reverse` method is `async` which it is not.

Another also, if messages will always be fetched in this order, maybe a default scope could have been used. Or a scope in general.

frontend/src/components/chat/ChatMessages.vue
```diff
@@ -39,11 +61,18 @@ export default {
       scroll = setTimeout(() => {
         const isUnderMessageHeight = offsetHeight + scrollTop + 0.5 < scrollHeight;
         scrollBehavior.smooth = isUnderMessageHeight;
+
+        if (scrollTop === 0 && wereMessagesAdded) {
+          loading.value = true;
+          offsetIndex.value++;
+          calculateScrollTop();
+        }
```
abasic, line 69: Nitpick:
```js
const isScrollOnTop = scrollTop === 0;
if (!isScrollOnTop || !wereMessagesAdded) return;

loading.value = true;
offsetIndex.value++;
calculateScrollTop();
```

abasic (changes requested):

backend/src/message/message.controller.ts
Lnola: How did that extra `await` sneak in? Anyway, I removed it and added `reverse` to a new line so it looks cleaner.

Regarding the default scope - messages will only be fetched in this order because this should be the only fetch messages used so I guess adding the default scope is fine 👍

frontend/src/components/chat/ChatMessages.vue
Lnola: I decided on this

```js
const loadMoreMessages = () => {
    loading.value = true;
    offsetIndex.value++;
    calculateScrollTop();
};
...

...
const isScrollOnTop = scrollTop === 0;
if (isScrollOnTop && wereMessagesAdded) loadMoreMessages();
```

abasic (approved):

sandje00 (approved):

PR #95, opened 2022-09-20
Fix: tags sort and accessibility

Lnola:
Add sort to tags - on fetch and select
Add close and plus icons to better differentiate selected and not selected tags


frontend/src/store/tagStore.js
```diff
@@ -12,6 +12,7 @@ const useTagStore = defineStore('tag', () => {
 
   const setTags = async () => {
     if (!tags.value.length) tags.value = await tagApi.getAll();
+    sortTags();
```
sandje00, line 15: Maybe this function call could also be part of the above `if` statement body?

Lnola: Doesn't mutate but doesn't need to sort if the array is empty

sandje00 (approved):

PR #98, opened 2022-09-20
Fix: remove active user hover

Lnola:
Add `hoverable` prop to ChiCard allowing to add hover effects when needed


frontend/src/components/sidebar/SidebarRooms.vue
```diff
@@ -9,6 +9,7 @@
           @click="handleSelect(element.id)"
           v-bind="element"
           :selected="roomStore.selectedRoom.id === element.id"
+          :hoverable="true"
```
sandje00, line 12: Is there a specific reason Boolean `true` value is passed this way instead of just listing `hoverable` as Boolean attribute?

Lnola: Fingers are truly faster than the mind 💅

sandje00 (approved):

PR #101, opened 2022-09-21
Fix Online Users per Room Count Inconsistency

sandje00:
THIS PR:

- Fix the issue of inconsistent online users per room count by ending socket connection on user logout

NOTE:

- Inconsistency in users per room count appears when a user logs out of the app, but has not closed the tab; in that situation the user is still visible within online users of the room and is still counted. In case of multiple logout / login within the same tab, the user is visible multiple times and the count is incorrect.


Lnola (commented):
this is my console after signing out

image

Lnola:
> this is my console after signing out
> 
> image

This has nothing to do with your PR since it also affects main, I'm pushing the change in a different PR

Lnola (approved):

PR #104, opened 2022-09-22
Feature: room interaction improvements

Lnola:
Add room id to socket and emit activity on disconnect
Fix multiple connection issues with emitting and disconnecting


frontend/src/socket/chat.js
```diff
@@ -5,7 +5,9 @@ import { useCookies } from '@vueuse/integrations/useCookies';
 
 class ChatService {
   createConnection() {
-    this.socket = io({ auth: { accessToken: useCookies('accessToken').get('accessToken') } });
+    this.socket = io('http://10.6.237.6:3001/', {
```
sandje00, line 8: Should this be an env variable?

backend/src/sockets/index.ts
```diff
@@ -7,7 +7,11 @@ import SocketActions from './actions';
 let io: Server;
 
 const socketConnection = (server: httpServer) => {
-  io = new Server(server);
+  io = new Server(server, {
+    cors: {
+      origin: '*',
```
sandje00, line 12: If we add origin to `env`, we could avoid `*`.

Lnola: we allow everyone to join so that's the reason for `*`, if we ever need it to be domain specific we can change it

frontend/src/socket/chat.js
Lnola: yep, it will be in a few seconds 😄

PR #116, opened 2022-09-28
Feature: move activity to navigation

Lnola:
- Remove users from sidebar and move activity change to room navigation
- Add animation to indicate new user joining/leaving the room
- Add `ChiModal` custom modal component
- Add `ChatActiveUsers` component that displays users currently active
- Make everything responsive


abasic (commented):
Screenshot 2022-09-28 at 18 09 16
The border around the user seems to be too dark. Maybe make it the same yellowish color.

PR #117, opened 2022-09-28
Feature: tags enhancement

Lnola:
Create custom plus icon - rotate it to create custom x icon and add the animation
Remove the icon from tags screen and rooms and add it as custom icon property


Lnola (commented):
Since we both worked on the issue but I raised the PR I cannot leave an approve, @sandje00 you will have to do that. I approve of the changes made by @sandje00, you can just review the part I did

sandje00 (approved):

PR #123, opened 2022-09-29
Fix: Rooms Seed Shift

sandje00:
THIS PR:

- Resolve room ids shift in rooms seed


Lnola (changes requested):
Can you revert this change and just change `The Room` to be the last room instead of the first 😄

Lnola (approved):

